PR #194, opened 2020-11-12
Fragmentate deposit market&fee account unspents

### LINKED ISSUE #184: Concurrent swaps: Fragment UTXOS
Each trade locks an utxo, therefore beside the size of the trade, the number of requests we can serve is linear with the number of nominal utxo we have.

In order to accommodate multiple concurrent trades at the same time, an easy solution is to fragment the utxos at deposit time, in many smaller utxos, so its easier to find utxo available for the trade.

This could be done either as external script or embedded into the daemon. 

This would means creating an ephemeral address for the user to deposit his UTXO, we would detect this, create a transaction that spends it and creates a number of smaller outputs versus the address of the given DepostiAddress rpc result.

The ratios of the amount of the outputs should reflect a scaled distribution, to be defined.

sekulicd:
- This adds the possibility for the operator to fragment unspents deposited to the market.
- DepositMarket RPC updated to accept new argument NumOfAddresses added to support fetching of multiple addresses for one market.
- This adds the possibility for the operator to fragment unspents deposited to the fee account.
- DepositFeeAccount RPC updated to accept new argument NumOfAddresses added to support fetching of multiple addresses for the Fee account.

Once operator invokes "depositmarket" command using CLI, an ephemeral address will be generated which needs to be funded with base and quote assets, once address is printed in console, crawler will be started which will watch for utxo's related ephemeral address.
When crawler notice that both assets are funded it will stop watching and it will craft transaction with 2 inputs (base & quote asset) and 23 outputs, 11 for base, 11 quote asset and one for fee which will be deducted from biggest base output.
11 outputs (base&asset) are calculated based on hardcoded fragment map(number of fragments utxo/percentage of main utxo) of  which is defined in task.
```
var fragmentationMap = map[int]int{
	1: 30,
	2: 15,
	3: 10,
	5: 2,
}
```

It closes #184 


tiero:
@sekulicd This is impossible to me to test with nigiri. I have a suspect is the "asset registry" used by chopsticks which is a json file.

Can you try it with a plain `docker-electrs` without Nigiri?

https://github.com/vulpemventures/nigiri/blob/master/resources/docker-compose-regtest-liquid.yml

cmd/tdex/depositfee.go
```diff
@@ -2,16 +2,38 @@ package main
...
+	MaxNumOfOutputs = 150
+)
+
 var depositfee = cli.Command{
 	Name:   "depositfee",
 	Usage:  "get a deposit address for the fee account used to subsidize liquid network fees",
 	Action: depositFeeAction,
+	Flags: []cli.Flag{
+		&cli.BoolFlag{
+			Name:  "no-fragment",
```
tiero, line 27: I was thinking that maybe better do the opposite, by default not doing anything.

Would modify this as just `--fragment`

cmd/tdex/depositfee.go
```diff
@@ -21,14 +43,228 @@ func depositFeeAction(ctx *cli.Context) error {
...
+		feeAmount,
+		net,
+		baseAssetKey,
+		[][]byte{randomWallet.BlindingKey()},
+	)
+	if err != nil {
+		log.Error(err)
+	}
+
+retry:
```
tiero, line 121: With the latest go-elements we are safe from blinding bugs, so I would remove

Issue #193, opened 2020-11-12
Synchronize database closing and crawler stopping


louisinger:
Hey! Problem with crawler & DB, any help is welcome :)

In my work-in-progress [PR](https://github.com/TDex-network/tdex-daemon/pull/179), I've implemented a system that open and close a mock database for **each** test. It allows to mock independent services such as crawlers. Thus, at the end of each test, we must (1) stop the crawler and (2) close the database. 

The problem is that, sometimes, the crawler still listening events AFTER the call of `stopObserveBlockchain` method. Maybe due to the [waiting step](https://github.com/louisinger/tdex-daemon/blob/issue_db_closed/pkg/crawler/service.go#L96) during quit ?? 
This leads to errors because the database is already closed when the crawler try to access it during event handling.

**How to reproduce**

1. clone the following branch: https://github.com/louisinger/tdex-daemon/tree/issue_db_closed
2. launch `make test`
_I set up some log to debug and see when the DB is open/close + log for crawler's event handler_

**Solution ??**

I would try to change the type of [quitChan](https://github.com/louisinger/tdex-daemon/blob/issue_db_closed/pkg/crawler/service.go#L42) from `chan int` to `chan func()`. This could let us to pass 'callback functions' to the `Stop` method. This function would be launch after the waiting step. I didn't try if it solve the problem, that's just an idea...

Guys, do you have any idea ?

PR #219, opened 2020-11-23
[Docker] Add integration testing for the docker container

louisinger:
**This PR adds a `cli_test.go` file in `cmd/tdex` which runs a set of cli commands inside tdexd containers**

TL;DR
- Add a new process in MakeFile: `containertest`
- Add CLI tests in `cli_test.go`: test for init and unlock the wallet + create, list and open a market

_I don't test all the methods because it could be a bit overkill in my opinion: we already test all of the CLI methods in others tests files. Please tell me if it is a good decision._

PS: Note that multiple tdexd containers are resistant to race errors --> could be the solution at the [former badger concurrency issue](https://github.com/TDex-network/tdex-daemon/issues/193). 

@sekulicd @tiero @altafan please review


altafan:
Guys, sorry for this soo late reply, but don't know why, i'm not able to follow this discussion from my laptop.. always getting a 500 error.
Not ideal to continue this from mobile, but however 😅...

With my latest PR, basically I tried to follow the first 3 points of the list given by dusan, which I essentially agree with.

It would be interesting to know if, using test suits, it would be possible to have isolated envs such that every suit could be ran into a t.Run block.

Running unit tests within t.Run blocks would be nice to speed up their execution. But since some of them, in particular those of the app layer, make use of mocked env vars this is currently not achievable because cause of race condition.

altafan:
One thing that i want to add to complete point 5 is that every state should be statically generated with a yml (also json file could be an option).

At the moment, there are mocks that are created "dynamically", in the sense they are based on hardcoded data that are used to create the actual mocks. Sometimes this is accomplished by using the same service that should be tested! It's fucking terrible trying to share a link from mobile, but take a look at how the trades are mocked in the app layer tests (test_util.go).

Ideally, instead, every unit test should start with a mocked state read from yml, json or whatever, execute a function, check the output and eventually destroy the state.

tiero:
I agree with @sekulicd

I would suggest to read also this post regarding testing https://threedots.tech/post/microservices-test-architecture and try to re-thing testing architecture from ground up

tiero:
Closing this in favor of #301

Issue #185, opened 2020-11-09
[Crawler] Starts handling blockchain events after unlock rpc


tiero:
Especially when restoring from a seed when we lost/deleted datadir﻿, better sto starts observing after the UnlockRPC, that means we already recovered from explorer and created the markets.

tiero:
This is pretty annoying indeed 

```
DEBU[0047] Start observing new account: 0               
DEBU[0047] Start observing new account: 0               
DEBU[0047] Start observing new account: 0               
DEBU[0047] Start observing new account: 0               
DEBU[0047] Start observing new account: 5               
DEBU[0047] Start observing new account: 5               
DEBU[0047] Start observing new account: 5               
DEBU[0047] Start observing new account: 5               
DEBU[0047] Start observing new account: 5               
DEBU[0047] ended syncing wallet                         
DEBU[0048] observe interval                             
WARN[0048] trying to check balance for fee account: wallet must be unlocked to perform this operation 
WARN[0048] trying to check fundings for market account 5: wallet must be unlocked to perform this operation 
WARN[0048] trying to check balance for fee account: wallet must be unlocked to perform this operation 
WARN[0048] trying to check fundings for market account 5: wallet must be unlocked to perform this operation 
WARN[0048] trying to check balance for fee account: wallet must be unlocked to perform this operation 
WARN[0048] trying to check balance for fee account: wallet must be unlocked to perform this operation 
WARN[0048] trying to check fundings for market account 5: wallet must be unlocked to perform this operation 
WARN[0048] trying to check fundings for market account 5: wallet must be unlocked to perform this operation 
WARN[0048] trying to check fundings for market account 5: wallet must be unlocked to perform this operation 
DEBU[0049] observe interval                             
WARN[0049] trying to check balance for fee account: wallet must be unlocked to perform this operation 
WARN[0049] trying to check balance for fee account: wallet must be unlocked to perform this operation 
WARN[0049] trying to check fundings for market account 5: wallet must be unlocked to perform this operation 
WARN[0049] trying to check balance for fee account: wallet must be unlocked to perform this operation 
WARN[0050] trying to check fundings for market account 5: wallet must be unlocked to perform this operation 
WARN[0050] trying to check balance for fee account: wallet must be unlocked to perform this operation 
WARN[0050] trying to check fundings for market account 5: wallet must be unlocked to perform this operation 
WARN[0050] trying to check fundings for market account 5: wallet must be unlocked to perform this operation 
[...]

PR #227, opened 2020-11-25
Set mnemonic in InitWallet for restoring wallet

### LINKED ISSUE #226: Error when restoring from seed mnemonic
I tried to restore from seed mnemonic and I get `rpc error: code = Internal desc = wallet must be unlocked to perform this operation`

Before this I already created a market, funded it, done couple of swaps. Then created another, but not funded.

Then I stopped, deleted datadir and re-run. When  call the InitWallet with the backupped seed mnemonic I get the error. I tried also with different machines, same behavior

### Stacktrace
```sh
tdex-daemon master ✅  export TDEX_CRAWL_INTERVAL=1000 
tdex-daemon master ✅  export TDEX_LOG_LEVEL=6
tdex-daemon master ✅  ./build/tdexd-linux-amd64 
INFO[0000] All 0 tables opened in 0s                    
INFO[0000] Replaying file id: 0 at offset: 0            
INFO[0000] Replay took: 1.677µs                         
INFO[0000] All 0 tables opened in 0s                    
INFO[0000] Replaying file id: 0 at offset: 0            
INFO[0000] Replay took: 2.344µs                         
INFO[0000] All 0 tables opened in 0s                    
INFO[0000] Replaying file id: 0 at offset: 0            
INFO[0000] Replay took: 2.215µs                         
DEBU[0000] start observe                                
INFO[0000] starting daemon                              
INFO[0000] trader interface is listening on :9945       
INFO[0000] operator interface is listening on :9000     
DEBU[0001] observe interval                             
DEBU[0002] observe interval                             
DEBU[0003] observe interval                             
DEBU[0004] observe interval                             
DEBU[0005] observe interval                             
DEBU[0006] observe interval                             
DEBU[0007] observe interval                             
DEBU[0008] observe interval                             
DEBU[0009] observe interval                             
DEBU[0010] observe interval                             
DEBU[0010] /Wallet/InitWallet                           
DEBU[0011] observe interval                             
DEBU[0012] observe interval                             
DEBU[0013] observe interval                             
DEBU[0013] start syncing wallet                         
DEBU[0014] observe interval                             
DEBU[0015] observe interval                             
DEBU[0016] observe interval                             
DEBU[0017] observe interval                             
DEBU[0018] observe interval                             
DEBU[0019] observe interval                             
DEBU[0020] observe interval                             
DEBU[0021] observe interval                             
DEBU[0022] observe interval                             
DEBU[0022] account 0 last derived external address 1    
DEBU[0023] observe interval                             
DEBU[0024] observe interval                             
DEBU[0025] observe interval                             
DEBU[0026] observe interval                             
DEBU[0027] observe interval                             
DEBU[0028] observe interval                             
DEBU[0029] observe interval                             
DEBU[0030] observe interval                             
DEBU[0031] observe interval                             
DEBU[0031] account 1 empty                              
DEBU[0032] observe interval                             
DEBU[0033] observe interval                             
DEBU[0034] observe interval                             
DEBU[0035] observe interval                             
DEBU[0036] observe interval                             
DEBU[0037] observe interval                             
DEBU[0038] observe interval                             
DEBU[0039] observe interval                             
DEBU[0040] observe interval                             
DEBU[0040] account 5 last derived external address 2    
DEBU[0041] observe interval                             
DEBU[0042] observe interval                             
DEBU[0043] observe interval                             
DEBU[0044] observe interval                             
DEBU[0045] observe interval                             
DEBU[0046] observe interval                             
DEBU[0047] observe interval                             
DEBU[0048] observe interval                             
DEBU[0049] account 6 empty                              
DEBU[0049] ended syncing wallet 
```﻿

### PR DESCRIPTION
**Pull request by tiero**
Before this commit was not possible to restore from seed

It closes #226 

Please review @altafan  @sekulicd 

(Why we didn't catch in tests it before?)

### DIFF
--- a/internal/core/application/wallet_service.go
+++ b/internal/core/application/wallet_service.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/hex"
 	"fmt"
+	"strings"
 
 	log "github.com/sirupsen/logrus"
 	"github.com/tdex-network/tdex-daemon/config"
@@ -150,6 +151,9 @@ func (w *walletService) InitWallet(
 			if err != nil {
 				return nil, err
 			}
+
+			// We unlock the vault setting the mnemonic in memory needed for restore
+			config.Set(config.MnemonicKey, strings.Join(mnemonic, " "))
 
 			feeLastDerivedIndex := getLatestDerivationIndexForAccount(ww, domain.FeeAccount, w.explorerService)
 			walletLastDerivedIndex := getLatestDerivationIndexForAccount(ww, domain.WalletAccount, w.explorerService)

### CODE REVIEW

File: internal/core/application/wallet_service.go [hunk 1]
```diff
@@ -151,6 +152,9 @@ func (w *walletService) InitWallet(
 				return nil, err
 			}
 
+			// We unlock the vault setting the mnemonic in memory needed for restore
+			config.Set(config.MnemonicKey, strings.Join(mnemonic, " "))
```
sekulicd, line 156: We could use vault method to unlock

```
if err = v.Unlock(passphrase); err != nil {
      return nil, err
}
```

sekulicd:
>  (Why we didn't catch in tests it before?)

Should  we write a test that InitWallet -> Unlock -> Derive and Fund address ->  Stop daemon -> delete data dir -> Start Daemon, Init and check if Address is derived or something

internal/core/application/wallet_service.go
altafan: +1

altafan:
>Should we write a test that InitWallet -> Unlock -> Derive and Fund address -> Stop daemon -> delete data dir -> Start Daemon, Init and check if Address is derived or something

There already is a test, but it's always skipped by the CI because it's not short.. [take a look](https://github.com/TDex-network/tdex-daemon/blob/master/internal/core/application/wallet_service_test.go#L95). I also suspect that this bug has been introduced when we modified the Vault by letting viper store the mnemonic, and since the test is always skipped, we didn't notice this before.

internal/core/application/wallet_service.go
altafan: Ok, I think I understood how we introudeced this bug. In this [commit](https://github.com/TDex-network/tdex-daemon/commit/200e6de2a169a5cfe303e0f5b386c76dfdcffc00#diff-872797e6e5246bb52e67bacdbd84b1ce2601e1e20a9e761454ceb4b9c59bf32f) we removed the Mnemonic filed from the Vault struct, and this is ok, but in the NewVault factory function we forgot to call config.Set, like done in Unlock, to store the plain text mnemonic into viper.
Shall we add this into NewVault instead of manually setting it as you've done?

tiero:
I also think that being global env variable, it just take one single test to manually set it and all consequent or parallel tests will have set already.

internal/core/application/wallet_service.go
tiero: Yes please do

tiero: Could you do in the fix with #185 maybe @altafan

PR #230, opened 2020-11-26
Trade complete time unix bug

### LINKED ISSUE #209: complete_time_unix of a completed Trade is earlier than accept_time_unix
I'm using `tdex-cli@beta` to make trades and with `tdex listswaps` I have the result:

```
{
	"swaps": [
		{
			"status": "COMPLETE",
			"amount_p": "10000000",
			"asset_p": "5ac9f65c0efcc4775e0baec4ec03abdde22473cd3cf33c0419ca290e0751b225",
			"amount_r": "182272727273",
			"asset_r": "2dd1d5aa89ce9be5656910208aa582ba4f5916491383fdef2832425f313b5290",
			"market_fee": {
				"asset": "5ac9f65c0efcc4775e0baec4ec03abdde22473cd3cf33c0419ca290e0751b225",
				"basis_point": "25"
			},
			"request_time_unix": "1605797092",
			"accept_time_unix": "1605797093",
			"complete_time_unix": "1605797092",
			"expiry_time_unix": "1605797212"
		}
	]
}
```﻿

### PR DESCRIPTION
**Pull request by sekulicd**
This fix bug related to setting trade completion status.

It closes #209

@tiero @altafan

### CODE REVIEW

File: internal/core/application/blockchain_listener.go [hunk 1]
```diff
@@ -87,49 +92,122 @@ func (b *blockchainListener) StopObserveBlockchain() {
...
+			ctx := context.Background()
+			if _, err := b.dbManager.RunTransaction(
+				ctx,
+				!readOnlyTx,
+				func(ctx context.Context) (interface{}, error) {
+					return nil, b.updateTrade(ctx, e)
+				},
+			); err != nil {
+				log.Warnf(
+					"update trade completion status failed %s\n",
```
tiero, line 146:
```suggestion
					"trying to update trade completion status %s\n",
```

internal/core/application/trade_service.go
```diff
@@ -374,6 +374,10 @@ func (t *tradeService) TradePropose(
 			trade.MarketFee = mkt.Fee
 			trade.MarketFeeAsset = mkt.FeeAsset
 
+			t.crawlerSvc.AddObservable(&crawler.TransactionObservable{
+				TxID: trade.TxID,
+			})
+
```
altafan, line 380: this should be done outside the closure, just after adding address observables

altafan:
With these changes, a trade could be retrieved by its txID if and only if the trader calls TradeComplete, because trade.Complete is the place where the TxID is set. Am I right?

1. We need to set the trade.TxID in trade.Accept
2. As the trade domain is designed, its fields are updated only by its methods, while info can be retrieved by both methods or public fields. I think this is a clear design, so I don't like so much updating the status in the closure as it's currently done within the bc listener.
3. I've got a proposal: what if we add a new status SETTLED that can come only after COMPLETE or ACCEPT-but-not-EXPIRED. This status would be set only by the bc listener once it detects the transaction included in the blockchain. We could create an ad hoc trade.Settle() method for this in the domain, that would be simply called by the listener.

tiero:
I agree on having an additional SETTLED status.

sekulicd:
> We need to set the trade.TxID in trade.Accept

TxID is added in trade.Complete  and trade.Accept in [here](https://github.com/sekulicd/tdex-daemon/blob/a8d86bb4f61e83f5ec641c139648c0fad3dfc64a/internal/core/domain/trade_service.go#L81)

> I've got a proposal: what if we add a new status SETTLED

I think that adding additional status would not give us more info about the entity but only add complexity, issue here is that we have two scenarios:
1. Client calls TradeComplete so that daemon can broadcast tx
2. Client itself broadcast tx
In both ways, we need to wait for tx to be confirmed on blockchain, only then (in the listener)  we can update the completion time, we have option to update trade.Complete method to support both flows, I generally don't prefer overcomplicated methods.
Maybe it would be easies to just update trade.AddBlocktime, rename it to trade.Confirm and add complete status in it.

tiero:
I think for now, let's just fix the bug we will think later about adding extra status.
let's the BC listener to setup the status COMPLETED and add txid etc..

If we pass from TradeComplete, we should only update the PSET, without touching the status and let the listener to put it that.

To differentiate this two paths we can have (only internally) two methods one Trade.Complete (that used by TradeComplete rpc to set the pset) and one Trade.Settle used by the listener

internal/core/domain/trade_service.go
```diff
@@ -172,12 +172,7 @@ func (t *Trade) Fail(swapID string, tradeStatus Status, errCode pkgswap.ErrCode,
 }
 
 // AddBlocktime sets the timestamp for a completed trade to the given blocktime.
-// If the trade is not in Complete status, an error is thrown
 func (t *Trade) AddBlocktime(blocktime uint64) error {
-	if !t.IsCompleted() {
```
altafan, line 165: weird, why you had to remove this even if in Settle(), before calling AddBlocktime, you're setting the status to `CompletedStatus`?

internal/core/domain/trade_service.go
```diff
@@ -72,6 +73,13 @@ func (t *Trade) Accept(
 	t.SwapAccept.Message = swapAcceptMsg
 	t.Timestamp.Accept = uint64(time.Now().Unix())
 	t.PsetBase64 = psetBase64
+
+	p, err := pset.NewPsetFromBase64(psetBase64)
+	if err != nil {
+		return false, err
+	}
+	t.TxID = p.UnsignedTx.TxHash().String()
```
altafan, line 81: can we add an ad hoc method in `transactionutil` pkg that does this? Just to avoid directly importing go-elements.

internal/core/domain/trade_service.go
sekulicd: [This](https://github.com/sekulicd/tdex-daemon/blob/0c75d87e25bb16f1e71dd4c361f28d0f67df2547/internal/core/domain/trade_service_test.go#L31) test was failing, should I revert and update test to use trade.Settle?

internal/core/domain/trade_service.go
```diff
@@ -137,15 +145,19 @@ func (t *Trade) Complete(psetBase64 string, txID string) (*CompleteResult, error
 		return &CompleteResult{OK: false}, nil
 	}
 
-	t.Status = CompletedStatus
 	t.SwapComplete.ID = swapCompleteID
 	t.SwapComplete.Message = swapCompleteMsg
 	t.PsetBase64 = psetBase64
-	t.TxID = txID
+	t.TxID = txHash
```
altafan, line 151: this can be just removed since we already set the TxID in trade.Accept()

internal/core/domain/trade_service.go
sekulicd: Done

internal/core/domain/trade_service.go
sekulicd: Done.

internal/core/domain/trade_service.go
sekulicd: Updated.

internal/core/domain/trade_service.go
```diff
@@ -73,12 +73,11 @@ func (t *Trade) Accept(
 	t.SwapAccept.Message = swapAcceptMsg
 	t.Timestamp.Accept = uint64(time.Now().Unix())
 	t.PsetBase64 = psetBase64
-
-	p, err := pset.NewPsetFromBase64(psetBase64)
+	txID, err := transactionutil.GetTxIdFromPset(psetBase64)
 	if err != nil {
-		return false, err
+		return false, nil
 	}
```
altafan: we can skip checking for error because we know that the pset is wall formed here

pkg/transactionutil/transactionutil.go
```diff
@@ -41,3 +42,12 @@ func NewFeeOutput(feeAmount uint64) []*transaction.TxOutput {
 		transaction.NewTxOutput(feeAsset, feeValue, feeScript),
 	}
 }
+
+func GetTxIdFromPset(psetBase64 string) (string, error) {
+	p, err := pset.NewPsetFromBase64(psetBase64)
+	if err != nil {
+		return "false", err
```
altafan, line 49: `return "",  err`

internal/core/application/blockchain_listener.go
```diff
@@ -169,7 +169,10 @@ func (b *blockchainListener) updateTrade(
 		ctx,
 		&trade.ID,
 		func(t *domain.Trade) (*domain.Trade, error) {
-			t.Settle(uint64(event.BlockTime))
+			err := t.Settle(uint64(event.BlockTime))
+			if err != nil {
+				return nil, err
+			}
```
altafan, line 175:
```
if err := t.Settle(uint64(event.Blocktime)); err != nil {
    return nil, err
}
```

pkg/transactionutil/transactionutil.go
tiero: cc @sekulicd

sekulicd: Done

internal/core/application/blockchain_listener.go
sekulicd: Done

PR #232, opened 2020-11-27
[Testing] Unlock wallet in wallet service test

### LINKED ISSUE #218: TestGenerateAddressAndWalletBalance needs vault to be unlocked to be run individually

louisinger:
**This PR unlocks the wallet before running unit tests in `TestGenerateAddressAndWalletBalance` & `TestSendToMany` in `wallet_service_test.go`**

- unlock wallet in TestGenerateAddressAndWalletBalance
- unlock wallet in TestSendToMany
- add error handling when status code != 200 for Faucet and Mint in `transaction.go`

it closes #218 

@tiero @altafan @sekulicd please review

### DIFF
--- a/internal/core/application/wallet_service_test.go
+++ b/internal/core/application/wallet_service_test.go
@@ -138,7 +138,7 @@ func TestWalletUnlock(t *testing.T) {
 	walletSvc, ctx, close := newTestWallet(dryLockedWallet)
 	defer close()
 
-	address, blindingKey, err := walletSvc.GenerateAddressAndBlindingKey(ctx)
+	_, _, err := walletSvc.GenerateAddressAndBlindingKey(ctx)
 	assert.Equal(t, domain.ErrMustBeUnlocked, err)
 
 	err = walletSvc.UnlockWallet(ctx, dryLockedWallet.password)
@@ -146,7 +146,7 @@ func TestWalletUnlock(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	address, blindingKey, err = walletSvc.GenerateAddressAndBlindingKey(ctx)
+	address, blindingKey, err := walletSvc.GenerateAddressAndBlindingKey(ctx)
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/internal/core/application/wallet_service_test.go
+++ b/internal/core/application/wallet_service_test.go
@@ -177,10 +177,16 @@ func TestGenerateAddressAndWalletBalance(t *testing.T) {
 	walletSvc, ctx, close := newTestWallet(dryWallet)
 	defer close()
 
+	err := walletSvc.UnlockWallet(ctx, dryWallet.password)
+	if err != nil {
+		t.Fatal(err)
+	}
+
 	address, _, err := walletSvc.GenerateAddressAndBlindingKey(ctx)
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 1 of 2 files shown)


internal/core/application/wallet_service_test.go
```diff
@@ -177,11 +177,17 @@ func TestGenerateAddressAndWalletBalance(t *testing.T) {
 	walletSvc, ctx, close := newTestWallet(dryWallet)
 	defer close()
 
+	err := walletSvc.UnlockWallet(ctx, dryWallet.password)
+	if err != nil {
+		t.Fatal(err)
+	}
+
```
altafan, line 184: the `dryWallet` used to be already unlocked, but since we moved the mnemonic into viper, it just lost this property.
Unlocking the wallet is an expensive operation in terms of time cost.
Do you think it's possible to set the mnemonic in newWalletTest at will and then unset it in the close function?

altafan: Otherwise, we have to skip this test in short mode.

tiero: I think we cannot set and unset the nev, otherwise we would need to do it for everyone.

Lets skip this in short mode.

altafan: We could indeed by simply calling `config.Set(config.MnemonicKey, "")` as the vault.Lock() method does, actually.

I think this is much better than skipping the test. I'm ok with skipping those that test the correctness of locking/unlocking and creating/restoring the wallet, or even the password change.
But all those functions that work only if the wallet is unlocked must not be skipped in my opinion, even more so if we can mock unlocked wallets.

Since we have different mocked wallets, like `dryWallet`, `emptyWallet`, etc, my concern was about setting into viper the correct mnemonic, but it's totally doable in `newTestWallet(w &mockedWallet)` through `config.Set(config.Mnemonic, w.mnemonic)`

tiero: ENV is gloabal, so if this test set it, it will be already "unlocked" for other test that came after, unless we "unset" which I don't know if is it possible

altafan: @louisinger looking at the tests of the internal/core/application pkg, we already have a "bad" test, [TestDepositMarket](../blob/master/internal/core/application/operator_service_test.go#L47), where we set the mnemonic to config. Because this mnemonic is never unset other wallets will result unlocked but the mnemonic used can potentially be incorrect.

Given this, It's necessary that every newTest<Service> function in test_util.go sets the mnemonic of whatever is the mockedWallet it's using, and also that every close function make sure to set the mnemonic back to an empty string.

This way we grant that every test works with an unlocked wallet, if necessary. It shouldn't be necessary to add other tests to those skipped.

louisinger: @altafan set the config leads to datarace as this is a global value. For instance `vaultRepository` need to fetch mnemonic from the config. However, as the tests runs concurrently: there is sometimes data race.
Moreover, when tests are running together they all modify the MnemonicKey of config so I don't know if it really solve the problem as the value change everytime we call `NewTest*` functions

tiero: That's true! This way now we export ENV vars like TDEX_NETWORK in the `make test` and we don't do programmatically. Viper is not safe for concurrent use.

PR #234, opened 2020-12-01
basic instrumentation added

sekulicd:
This adds basic instrumentation, it prints basic process info like memory allocation and a number of goroutines.
On server stop event all default prometheus stats are printed to a file, this can potentially help to see basic info before server is stopped.

@tiero @altafan please review if you think that this can be usefull.

### DIFF
--- a/cmd/tdexd/main.go
+++ b/cmd/tdexd/main.go
@@ -115,13 +115,17 @@ func main() {
 
 	log.Info("starting daemon")
 
-	statsDir := filepath.Join(config.GetString(config.DataDirPathKey), "stats")
-	ctx, cancelStats := context.WithCancel(context.Background())
-	stats.EnableMemoryStatistics(
-		ctx,
-		config.GetDuration(config.StatsIntervalKey)*time.Second,
-		statsDir,
-	)
+	var cancelStats context.CancelFunc
+	if log.GetLevel() >= log.DebugLevel {
+		statsDir := filepath.Join(config.GetString(config.DataDirPathKey), "stats")
+		var ctx context.Context
+		ctx, cancelStats = context.WithCancel(context.Background())
+		stats.EnableMemoryStatistics(
+			ctx,
+			config.GetDuration(config.StatsIntervalKey)*time.Second,
+			statsDir,
+		)
+	}
 
 	defer stop(
 		dbManager,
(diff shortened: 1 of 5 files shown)


pkg/stats/stats.go
```diff
@@ -0,0 +1,102 @@
...
+		toGigabytes(bytesHeapAllocated),
+		countMalloc,
+		countFrees,
+	)
+}
+
+// DumpPrometheusDefaults write default Prometheus metrics to a file
+func DumpPrometheusDefaults() error {
+	file, err := os.OpenFile(
+		fmt.Sprintf("stats_%v", time.Now().Format(time.RFC3339)),
```
tiero, line 73: Let's write it in the data dir, in a folder like `logs` or `stats`

sekulicd: Done

pkg/stats/stats.go
```diff
@@ -0,0 +1,109 @@
...
+func PrintMemoryStatistics() {
+	var memStats runtime.MemStats
+	runtime.ReadMemStats(&memStats)
+
+	bytesTotalAllocated := memStats.TotalAlloc
+	bytesHeapAllocated := memStats.HeapAlloc
+	countMalloc := memStats.Mallocs
+	countFrees := memStats.Frees
+
+	log.Infof(
```
tiero, line 65: Debugf maybe?

pkg/stats/stats.go
```diff
@@ -0,0 +1,109 @@
...
+
+	writer.Flush()
+	file.Close()
+
+	return nil
+}
+
+// PrintNumOfRoutines prints number of go routines currently running
+func PrintNumOfRoutines() {
+	log.Infof("Num of go routines: %v\n", runtime.NumGoroutine())
```
tiero, line 108: Debugf ?

pkg/stats/stats.go
sekulicd: Done

pkg/stats/stats.go
sekulicd: Done

cmd/tdexd/main.go
```diff
@@ -113,11 +115,20 @@ func main() {
 
 	log.Info("starting daemon")
 
+	statsDir := filepath.Join(config.GetString(config.DataDirPathKey), "stats")
+	ctx, cancelStats := context.WithCancel(context.Background())
+	stats.EnableMemoryStatistics(
```
tiero, line 120: shoudl this be something we enable in sort of debug mode or we want it always on?

sekulicd: If we only use Debugf then there is no sense in waiting for resources, so yes I will add this check.

sekulicd: Done

PR #243, opened 2020-12-03
Crawler performance improvement

sekulicd:
This improves crawler parallelization and CPU utilisation.
Before this crawler was creating one goroutine for each address that needs to be crawled(in one crawl interval) and it was placing Event in unbuffered channel (same for errors that were sent to err chan), on the other side blockchain listener was processing events from unbuffered channel, meaning one event at a time.
In other words many goroutines were waiting for the event chan to be freed.
Now it is added a buffered channel for both event and error with size equals to constants:
eventQueueMaxSize = 100
errorQueueMaxSize = 10
this could be something to watch over and tune it as needed.
Also blockchain listener is slightly updated to that it is processing each event in its own routine.

This drastically improves performances and now it is expected that server won't be crushing.
Note that this is not tested with realistic load, with max number of markets and big number of addresses.


tiero:
That seems a nice improvement indeed. 

@altafan thoughts?

internal/core/application/blockchain_listener.go
```diff
@@ -86,51 +90,53 @@ func (b *blockchainListener) StopObserveBlockchain() {
...
-			!readOnlyTx,
-			func(ctx context.Context) (interface{}, error) {
-				return nil, b.updateUnspentsForAddress(ctx, unspents, e.Address)
-			},
-		); err != nil {
-			log.Warnf("trying to update unspents for address %s: %s\n", e.Address, err.Error())
-			break
-		}
+	for crawlerEvent := range b.crawlerSvc.GetEventChannel() {
+		go func(event crawler.Event) {
```
altafan, line 94: let's give this function a name such as `handleEvent` or `eventHandler` or something better, what do you think?

tiero:
Solve conflicts

internal/core/application/blockchain_listener.go
sekulicd: Done

altafan:
I tested this by running the daemon... I can say that this is nice, I didn't saw any of those annoying warnings for accounts not funded after a trade anymore. Sometimes I see the message `fee account funded. Trades can be served` even if no warning was printed before, but not so concerned about it. It just means that there's still space for more improvements.

LGTM

PR #244, opened 2020-12-03
Update rpc to newer version of protobuf & Improve MarketPrice

### LINKED ISSUE #218: TestGenerateAddressAndWalletBalance needs vault to be unlocked to be run individually

### LINKED ISSUE #238: Update MarketPrice rpc to follow the spec
The latest trade specification has an asset field, that will allow to make a preview based on both base or quote asset﻿

altafan:
This updates the version of the tdex-protobuf dep to a newer version (not the latest because it requires other changes out of this scope).

Some of the existing RPCs had to be changed in accordance with the new specifications. In particular, the `MarketPrice` rpc of the trade service now can calculate the preview amount by providing also the quote asset. Prior to this instead, it was expecting only the amount of base asset, thus calculating exclusively quote asset amounts.

This also refactors the existing tests for the various RPC of the application services in a way that before each of them, the db, repos, etc are initialized with mocked data, and also with a mnemonic set to ENV through config to simulate an unlocked wallet if necessary. After a test terminates, all things are unset so that the following don't have corrupted data or whatsoever that can alter its behavior.

For every RPC there should be 2 tests, one for valid cases and one for failing cases (marked with `TestFailingXXXX`).
Every testcase that was previously run within t.Run, now is an entry in an array of test cases in a `[]struct{}` slice usually called `tests`. Basically, all the existing test cases were kept, except for some of them that weren't testing anything particularly interesting.

The test `TestDepositMarketWithCrawler` has been removed since it's not a unit test in the sense that requires more than just the single operator service of the app layer to work as expected. And also, it wasn't either working properly, but since it was marked as not Short, none of us has noticed it. 

The use of t.Run has been removed for now because we need to properly understand how to parallelize testing so that we don't encounter race conditions. So I'm expecting tests will be subject of another refactor in the future, but we need to agree on a standard design on how to write them first.

This closes #238.
This closes #218.

Please @tiero @louisinger @sekulicd review this.


internal/core/application/trade_service.go
```diff
@@ -326,13 +326,14 @@ func (t *tradeService) TradePropose(
...
 
-			if !isValidTradePrice(swapRequest, tradeType, previewAmount) {
+			if !isValidTradePrice(swapRequest, tradeType, mkt) {
 				trade.Fail(
 					swapRequest.GetId(),
 					domain.ProposalRejectedStatus,
 					pkgswap.ErrCodeInvalidSwapRequest,
 					"bad pricing",
 				)
+				swapFail = trade.SwapFailMessage()
```
tiero, line 336: we didnt catched early bacuase we didnt had a failing test case for this right

internal/core/application/trade_service.go
```diff
@@ -762,20 +764,20 @@ func getBalanceByAsset(unspents []domain.Unspent) map[string]uint64 {
...
-func calcPreviewAmount(market *domain.Market, tradeType int, amount uint64) uint64 {
+func calcPreviewAmount(market *domain.Market, tradeType int, amount uint64, asset string) uint64 {
+	var price decimal.Decimal
+	if asset == market.BaseAsset {
+		price = market.QuoteAssetPrice()
+	} else {
+		price = market.BaseAssetPrice()
+	}
+
+	chargeFeesOnTheWayIn := asset == market.BaseAsset
```
tiero, line 775: mmm i donth think chargeFeesOnTheWayIn means this

internal/core/application/trade_service.go
```diff
@@ -885,7 +922,24 @@ func previewFromFormula(
...
+	if tradeType == TradeSell {
+		amount = swapRequest.AmountP
+	}
+
+	_, previewAmount, _ := getPriceAndPreviewForMarket(
+		marketUnspents,
+		market,
+		tradeType,
+		amount,
+		market.BaseAsset,
```
tiero, line 936: it shoudl switch based on the way of the tradeType

internal/core/application/trade_service.go
```diff
@@ -326,13 +326,14 @@ func (t *tradeService) TradePropose(
...
-			if !isValidTradePrice(swapRequest, tradeType, previewAmount) {
+			if !isValidTradePrice(swapRequest, tradeType, mkt) {
 				trade.Fail(
 					swapRequest.GetId(),
 					domain.ProposalRejectedStatus,
 					pkgswap.ErrCodeInvalidSwapRequest,
 					"bad pricing",
 				)
+				swapFail = trade.SwapFailMessage()
 				return trade, nil
```
tiero, line 337: We should store the fee in the collected trade domain isn it_

internal/core/application/trade_service.go
```diff
@@ -162,18 +170,24 @@ func (t *tradeService) GetMarketPrice(
 		return nil, err
 	}
 
-	price, previewAmount, err := getPriceAndPreviewForMarket(unspents, mkt, tradeType, amount)
+	price, previewAmount, err := getPriceAndPreviewForMarket(unspents, mkt, tradeType, amount, asset)
```
tiero, line 173: shouldt we pass the fee to calculate proper preview

tiero (changes requested):
Since you updated the proto stubs, also the `cmd/tdex` should be updated to the new request repsonse. @altafan

internal/core/application/trade_service.go
altafan: do you have any suggestion on how to rename this? I kept the same nomenclature of the marketmaking package since calcProposeAmount is the equivalent of the inGivenOut, while calcExpectedAmount corresponds to OutGivenIn if we try to make a comparison between strategy methods.

Also, I'm thinking that maybe these functions could be moved to the market domain in the future, wrapped by one method that takes care of calculating the price depending on its current strategy. What do you think? Eventually, this won't be done here but in another PR, after opening a dedicated ticket.

internal/core/application/trade_service.go
altafan: the fact is that even depending on the trade type, it still can be possible that the trader fixed the quote asset amount to get the corresponding base asset one when getting the preview. For this reason, the daemon cannot be sure of the way it should verify prices, so I decided to always verify that the LBTC/ASSET price is valid.

internal/core/application/trade_service.go
altafan: yes, but it's not in the scope of this PR. Don't want to add other changes to fix the ReportMarketFee rpc.

internal/core/application/trade_service.go
altafan: we already include fees for both balanced and pluggable strategies.

internal/core/application/trade_service.go
tiero: in the sens we should change the name of the variable to not make confusuon

tiero: The idea od being a pkg is that other people can make their own and we use them, as long they implement the same interface.

About the name I guess the coorect way is : "isBaseAsset" beaccuse is what is actually is

internal/core/application/trade_service.go
tiero: but since amount switches based on the trade type, also the asset should follow.

internal/core/application/trade_service.go
altafan: I'm saying that, for now, we decide at the application level if the preview amount is calculated from a formula (balance strategy), or from a price (pluggable strategy). I think that it would be fine if we had a SpotPrice method in the market domain so that we move this logic to the domain level.

internal/core/application/trade_service.go
altafan: the amount does not switch based on the trade type but on the asset passed to the MarketPrice. If I want to make a BUY passing an LBTC amount, the USDT one will be calculated and vice-versa.

For this reason, with the last commits, I added the validation for both the 2 predicted assets of a swap. I mean that if fixing the AssetP, the AmountR will be calculated based on the current price, and used to create a range in which the actual swap.AmountR must be included.

The same is done by fixing the AssetP though.

If both the 2 amounts of the swap are not included in the ranges, then the swap is invalid because of `bad pricing`. Walking the "wrong" way, which means fixing the amount that was actually calculated to predict the one that was actually fixed, does not grant to get back an amount that is enough close to the one of the swap. This depends essentially on the slippage % used to create the amount range.

For this reason, it's enough that one of the 2 ways result in a positive result to be sure that the amounts of the swap satisfy the current market price.

tiero:
> Since you updated the proto stubs, also the `cmd/tdex` should be updated to the new request repsonse. @altafan

cc @altafan

tiero (changes requested):
Other thing i would do is to check the amount coming in the trade in the pluggable strategy is not greater than the balances we have. Now the only way it to make an actual trade to know, when we should be consistent in the MarketPrice between two strategies

altafan:
@tiero I added a --num_of_addresses flag to the `depositfee` and `depositmarket` CLI commands, and also add the checkpoints to make sure that an amount passed to the MarketPrice is not or does not have a counter amount bigger than its corresponding balance. This is done also when validating the price of a trade proposal.

internal/core/application/trade_service.go
```diff
@@ -734,21 +735,62 @@ func getPriceAndPreviewForMarket(
...
+		amount,
+		asset,
+	)
+	if err != nil {
+		return Price{}, 0, err
+	}
+
+	if tradeType == TradeBuy {
+		if asset == market.QuoteAsset && previewAmount >= baseAssetBalance {
+			return Price{}, 0, errors.New("preovided amount is too big")
```
tiero, line 785:
```suggestion
			return Price{}, 0, errors.New("given amount is too big")
```

altafan: I used the same message of the `marketmaking` pkg.

PR #252, opened 2021-02-16
Optimize memory compsumption

### LINKED ISSUE #249: Optimize memory consumption on disk
After 13days of utilization of a live instance of the tdex-daemon, it eated up the whole 8GB of the disk on aws ec2 instance. This beacuse badger do not performs automatic garbage collection and because is not tweaked to "eat" less space as possible.

- [] Garbage collection
- [] Optimized badger configuration 
- [] use binary serialization instead of JSONEncode

References:

https://dgraph.io/docs/badger/get-started/#garbage-collection
https://dgraph.io/docs/badger/get-started/#memory-usage
﻿

altafan:
This fixes #249:
* Add garbage collection
* Change binary serialization instead of JSON
* Tweaks badger configuration to avoid memory-mapping log files.

Moving to binary serialization drastically improves memory usage.

I tried to run 3 different daemons with the same configuration and by repeating the same operations:
* derive 5 fee addresses and fund each of them with 10 utxos (50 in total)
* derive 5 addresses for a market and fund the market with 1 utxo per each address

The first daemon was the master version and the datadir, overall, ended occupying 4.7MB on disk.
The second daemon had all the changes but the binary serialization and its datadir was 3.7MB.
The last daemon had all the changes and took only 315KB on the disk.

Please @tiero, review this.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/btcsuite/btcd v0.20.1-beta
 	github.com/btcsuite/btcutil v1.0.2
 	github.com/desertbit/timer v0.0.0-20180107155436-c41aec40b27f // indirect
+	github.com/dgraph-io/badger v1.6.2
 	github.com/dgraph-io/badger/v2 v2.2007.2
 	github.com/gogo/protobuf v1.2.1
 	github.com/google/uuid v1.1.2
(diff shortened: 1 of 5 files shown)


tiero:
We may want to use compression as well?

tiero (approved):

PR #255, opened 2021-02-19
Experimental elements

### LINKED ISSUE #240: Plug Elements node as source for unspents
Add EXPERIMENTAL env var to switch the crawler&cache ON and default to elements node.

A (attempt) starting point could be found here  https://github.com/tiero/tdex-daemon/commit/4d00a8b2c4b730fc30b0f3f1c9922c3fd0c02dbe

### LINKED ISSUE #248: Remove config dependency from /pkg
Currently packages in /pkg are dependent to config.
One reason why this is problematic is that  [init](https://github.com/TDex-network/tdex-daemon/blob/474c38aa1123f231bdcff8fe477ffb09ec03ee4d/config/config.go#L58) is invoked in other projects where we could use tdex-daemon pakcakges.

altafan:
This adds the Elements implementation of the Explorer interface that connects to a Elements node and calls its RPCs under the hood.
Started from @tiero's work, the original elements service became an implementation of the Explorer interface, just like esplora. Therefore, the `pkg/explorer` package has been refactor so that it has ONLY interfaces defined, and subfolders `elements` and `esplora` that contain their respective implementations. 

A new `TDEX_EXPERIMENTAL` flag has been added along with `TDEX_ELEMENTS_RPC_HOST | TDEX_ELEMENTS_RPC_PORT | TDEX_ELEMENTS_RPC_USER | TDEX_ELEMENTS_RPC_PASSWORD` to configure the connection to the node.

NOTE: You can test the correctness of this explorer by adding the above env vars to `make run` or by building the project and exporting them along with those recommended. We don't have integration tests to validate the inter-operation of differents daemon's components though. Therefore, only unit tests has been added for the new Explorer implementation.

BONUS: I increased the STATS interval from 10 secs to 10 mins because it was really annoying in the logs so far.

Closes #240.
Also closes #248.

Please @tiero review this.


config/config.go
```diff
@@ -50,6 +50,17 @@ const (
 	EnableProfilerKey = "ENABLE_PROFILER"
 	// StatsIntervalKey defines interval for printing basic tdex statistics
 	StatsIntervalKey = "STATS_INTERVAL"
+	// ExperimentalKey defines wether to use an Elements node as the daemon's
+	// block explorer instead of the default Esplora service.
+	ExperimentalKey = "EXPERIMENTAL"
+	// ElementsRPCHostKey is the RPC host of the Elements node. Experimental ONLY.
+	ElementsRPCHostKey = "ELEMENTS_RPC_HOST"
```
tiero, line 57: I would do different here
```suggestion
	ElementsJsonRpcEndpointKey = "ELEMENTS_JSON_RPC_ENDPOINT"
```

Winch would be in the canonical URL form of `protocol://user:pass@host:port`

The golang url package provides `url.Parse` to handle this

config/config.go
```diff
@@ -50,6 +50,17 @@ const (
 	EnableProfilerKey = "ENABLE_PROFILER"
 	// StatsIntervalKey defines interval for printing basic tdex statistics
 	StatsIntervalKey = "STATS_INTERVAL"
+	// ExperimentalKey defines wether to use an Elements node as the daemon's
+	// block explorer instead of the default Esplora service.
+	ExperimentalKey = "EXPERIMENTAL"
```
tiero, line 55: I think here we could do it a simpler way to decrease number of ENV, I would make `ElementsJsonRpcEndpointKey` and `ExplorerEndpointKey` exclsive, chechink this in the Validate method (ie. throw Fatal error if both are given or neither of two)

tiero: therefore dropping the experimental, wich could just be documented in the GoDoc comment on top of Elements key

pkg/explorer/coinselection.go
```diff
@@ -1,122 +1,55 @@
...
-		select {
-		case err := <-chErr:
-			close(chErr)
-			close(chUnspents)
-			return nil, err
-		case unspentsForAddress := <-chUnspents:
-			unspents = append(unspents, unspentsForAddress...)
+// SelectUnspents performs a coin selection over the given list of Utxos and
+// returns a subset of them of type targetAsset to cover the targetAmount.
+// A list of blindKeys is required in case any of them needs to be unblinded.
```
tiero, line 11: is the comment still accurate? ` A list of blindKeys ...`

pkg/explorer/coinselection.go
```diff
@@ -1,122 +1,55 @@
...
-		coins = nil
-		err = fmt.Errorf(resp)
-		return
-	}
-
-	var witnessOuts []witnessUtxo
-	err1 = json.Unmarshal([]byte(resp), &witnessOuts)
-	if err1 != nil {
+	selectedUtxos := make([]Utxo, 0)
+	if len(indexes) > 0 {
```
tiero, line 36: Invert the check without using else branching eg. `if len(indexed) == 0` > return error

pkg/explorer/elements/service.go
```diff
@@ -0,0 +1,90 @@
...
+func NewService(host string, port int, user, passwd string) (explorer.Service, error) {
+	client, err := NewClient(host, port, user, passwd, false, 30)
+	if err != nil {
+		return nil, err
+	}
+
+	return &elements{client}, nil
+}
+
+func (e *elements) mine() (string, error) {
```
tiero, line 24: I woudl add a comment to makes this clear as regtest-only util

pkg/explorer/elements/service.go
```diff
@@ -0,0 +1,90 @@
...
+	"fmt"
+
+	"github.com/tdex-network/tdex-daemon/pkg/explorer"
+)
+
+type elements struct {
+	client *RPCClient
+}
+
+// NewService returns and Elements based explorer
```
tiero, line 14: Being a public package (ie. GoDoc self generates documentation out of it) I would document a bit more the public methods (what they expect and what they actually do)

ie. here we are opnening a insecure no-TLS connection by default, that's not obvious for end user

and so on in other methods :pray:

pkg/explorer/elements/transaction.go
```diff
@@ -0,0 +1,200 @@
...
+		info := data[0].(map[string]interface{})
+		txids := info["txids"].([]interface{})
+		confirmations := int(info["confirmations"].(float64))
+
+		txs := make([]explorer.Transaction, 0, len(txids))
+		chTxs := make(chan explorer.Transaction)
+		chErr := make(chan error, 1)
+
+		for _, txid := range txids {
+			go e.getTxDetails(txid.(string), confirmations, chTxs, chErr)
```
tiero, line 95: add TODO: with blinders support no need to get the tx

config/config.go
altafan: consider that atm we set a default value for the Explorer endpoint [here](../blob/master/config/config.go#L65), so technically what you say it's not doable unless we remove that default value.

altafan: @tiero instead of making them mutually exclusive, I opted for being enough to set a TDEX_ELEMENTS_RPC_ENDPOINT to set the Elements explorer instead of the default Esplora one.

tiero: Ok so if both set, Elements take precedence? That's ok too (unless we remove the default)

pkg/explorer/elements/transaction.go
```diff
@@ -29,7 +29,7 @@ func (e *elements) IsTransactionConfirmed(txid string) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	return data["confirmed"].(bool), nil
+	return data["confirmations"].(float64) > 0, nil
```
tiero, line 32: usually better always checking the second return value `ok` when casting an `interface{}`

```
nConf, ok := data["confirmations"].(float64)
if !ok {
....
}
```

altafan: the problem was on the field, `confirmations` always exists in the response

.github/workflows/go.yml
```diff
@@ -30,4 +30,6 @@ jobs:
         run: go get -v -t -d ./...
 
       - name: Test
-        run: make test
+        run: |
```
tiero, line 33: would you add test step for both in the CI?

cmd/tdexd/main.go
```diff
@@ -63,7 +66,16 @@ func main() {
 	marketRepository := dbbadger.NewMarketRepositoryImpl(dbManager)
 	tradeRepository := dbbadger.NewTradeRepositoryImpl(dbManager)
 
-	explorerSvc := explorer.NewService(config.GetString(config.ExplorerEndpointKey))
+	var explorerSvc explorer.Service
+	if rpcEndpoint := config.GetString(config.ElementsRPCEndpointKey); len(rpcEndpoint) > 0 {
```
tiero, line 70: I think this should live in the Validate method of the config package

config/config.go
```diff
@@ -33,7 +33,7 @@ const (
 	// BaseAssetKey is the default asset hash to be used as base asset for all markets. Default is LBTC
 	BaseAssetKey = "BASE_ASSET"
 	// CrawlIntervalKey is the interval in milliseconds to be used when watching the blockchain via the explorer
-	CrawlIntervalKey = "CRAWL_INTERVAL"
+	CrawlIntervalKey = "CRAWL_INTERVAL_IN_MILLISECS"
```
tiero, line 36: I prefer to have shorter ENV variable here. The GoDoc is the place where we point the users to know about the EN vars, we already specify is in milliseconds there

.github/workflows/go.yml
altafan: do you mean running also `make integrationtest`?

cmd/tdexd/main.go
altafan: do you mean a `config.GetExplorerService()` getter, for instance?

tiero: Like we do for the SSL check. A specific getter also could be useful

altafan: the correctness of the Elements endpoint is checked into the `elements.NewService()` factory function... furthermore, if we add a `config.GetExplorer()` getter that takes care of instantiating the correct explorer depending on the env vars, we would still check the correctness of the endpoint at config level, but without repeating the same checks twice.

.github/workflows/go.yml
tiero: no running a tets with elements and one with electrs

altafan: this could be done in app level tests, like testing every operator/trade/wallet methods with esplora and with elements explorers.. sounds good?

tiero:
Solve conflicts @altafan

PR #258, opened 2021-02-23
get block height

sekulicd:
This adds new method, GetBlockHeight, to the explorer package.
It returns block tip of the blockchain.

@tiero please review.

### DIFF
--- a/pkg/explorer/block.go
+++ b/pkg/explorer/block.go
@@ -1,0 +1,29 @@
+package explorer
+
+import (
+	"fmt"
+	"github.com/tdex-network/tdex-daemon/pkg/httputil"
+	"net/http"
+	"strconv"
+)
+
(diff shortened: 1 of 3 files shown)


tiero (approved):

pkg/explorer/block.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+func (e *explorer) GetBlockHeight() (int, error) {
+	url := fmt.Sprintf(
+		"%v/blocks/tip/height",
+		e.apiUrl,
+	)
+	status, resp, err := httputil.NewHTTPRequest("GET", url, "", nil)
+	if err != nil {
+		return 0, err
```
altafan, line 17: I would return -1 in case of error
```suggestion
		return -1, err
```

PR #259, opened 2021-02-24
Crawler and BlockchainListener improvements

### LINKED ISSUE #256: Improve Crawler observation
At the moment, the **crawler** makes use of a `observeAll()` function that periodically loops over the list of observables and calls their `observe()` methods. This means that the GetEventChannel() is somehow "sync" because every X second (potentially) a lot of info needs to be processed with several `observe()` invoked one right after the other. Furthermore, at certain point this can causes the explorer to ban the daemon because it makes too many requests in very little time.

One solution could be to not have an `observeAll()` function at all.. in `AddObservable()`, instead, we could directly start a go routine that periodically calls `observe()`. The crawler should have a map of every go routines to easily find and stop them whenever `RemoveObservable()` is called. This way we would have observations distributed in time instead of doing all of them in a very short time-window of a for loop.

### LINKED ISSUE #257: Crawler should be used only by blockchain listener
At the moment, there are several components of the application level that expect to use a crawler interface (only for adding observables to it). Ideally, ONLY the BlockchainListener one should interface with it instead.

One solution could be to wrap the crawler's AddObservable invocation into a BlockchainListener's method, so that app level services expect to use a bcListener for adding address/txs to the watchlist.

altafan:
This contains a little refactor of the crawler pkg and also of the BlockchainListener service of the application level.

Before this, the crawler was using a single ticker for all its observables, now, on the contrary, every observable has its own ticker (every ticker has the same duration still - refers to CRAWL_INTERVAL env var).
The usage of its inner channels has been fixed:
* the original `quitChan` has been removed
* `errChan` is closed when stopping the crawler
* a special event `QuitEvent` is now sent over the `eventChan` to notify the client (listening for events with `GetEventChannel()`) that no more events will be sent over the channel, which can be considered closed. In fact, this happens ONLY when stopping the crawler. Since the crawler manages many go-routines (one per each observable), the `eventChan` can't be safely closed, generating a race condition. This is the reason why the channel is not closed but a close signal is sent over it instead.
* a new `observableHandler` type has been added which takes care of starting and stopping the observation go-routine. When stopping the crawler, every observableHandler is stopped, and only after that the `CloseEvent` is sent over the `eventChan`. This way it's not possible that some other events are read from that channel after `CloseEvent`.

Before this, services were making directly use of the crawler service (basically only for adding addresses and txs). Starting from this, instead, the BlockchainListener has been refactored a bit so that other app services use it instead of the crawler_
* existing start/stop methods have been renamed `StartObservation()` and `StopObservation()`
* new public methods for adding/removing addresses or txs to the crawler's observations
* prevent adding observables to the crawler before it has been started (this was actually happening and we weren't aware of this problem yet) - the bc listener now has a pendingObservables field which is filled in such situation and emptied right after starting the service.

BONUS: before this, every app service was making a blocking call for adding observables; now these kind of operations are ALL non-blocking (make use of go-routine) except for only one situation: when creating the wallet service we already check if it can be restored from storage in [`newWalletService`](https://github.com/altafan/tdex-daemon/blob/crawler-async/internal/core/application/wallet_service.go#L82). In this case the observation phase must be blocking because the wallet shouldn't be considered initialized BEFORE the listener has started tracking all wallet addresses.

This closes #256.
This closes #257.

Please @tiero @sekulicd  review this.

### DIFF
--- a/internal/core/application/blockchain_listener.go
+++ b/internal/core/application/blockchain_listener.go
@@ -3,6 +3,7 @@ package application
 import (
 	"context"
 	"sync"
+	"time"
 
 	log "github.com/sirupsen/logrus"
 	"github.com/tdex-network/tdex-daemon/config"
@@ -16,19 +17,26 @@ const readOnlyTx = true
 
 //BlockchainListener defines the needed method sto start and stop a blockchain listener
 type BlockchainListener interface {
-	ObserveBlockchain()
-	StopObserveBlockchain()
+	StartObservation()
+	StopObservation()
+
+	StartObserveAddress(accountIndex int, addr string, blindKey []byte)
+	StopObserveAddress(addr string)
+
+	StartObserveTx(txid string)
+	StopObserveTx(txid string)
 }
 
 type blockchainListener struct {
-	unspentRepository domain.UnspentRepository
-	marketRepository  domain.MarketRepository
-	vaultRepository   domain.VaultRepository
-	tradeRepository   domain.TradeRepository
-	crawlerSvc        crawler.Service
-	explorerSvc       explorer.Service
-	dbManager         ports.DbManager
-	started           bool
+	unspentRepository  domain.UnspentRepository
+	marketRepository   domain.MarketRepository
+	vaultRepository    domain.VaultRepository
+	tradeRepository    domain.TradeRepository
+	crawlerSvc         crawler.Service
+	explorerSvc        explorer.Service
+	dbManager          ports.DbManager
+	started            bool
+	pendingObservables []crawler.Observable
 	// Loggers
 	feeDepositLogged    bool
 	feeBalanceLowLogged bool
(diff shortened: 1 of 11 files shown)


sekulicd:
@tiero @altafan as discussed offline i am sharing my thoughts with you.
As described in issue #256 "... this can causes the explorer to ban the daemon because it makes too many requests in very little time." 
If i understood correctly, we are trying to prevent 'max num of calls exceeded in interval x' which is set on server side(elements). 
Proposed solution, with separate ticker for each address, doesnt guarantee that we wont exceed max num of call in specific period. 
Number of addresses most probably will increase over time and lets say we have 10K of addresses and that rate limit set on server side(elements) is 100calls in 1 sec, if we have ticker for each address it can happen that in one second there can be more calls than 100(depends on interval value, if it is large enough it wont happen, but in this case i think it must be 100seconds if i am not wrong, just want to explain my point).

Possible solution:
1. We find out what is rate limit on server side
2. We define worker pool of X number of go routines(X=num of calls that server can handle) and interval Y (Y=server interval) and we scan X number of addresses in one interval
We need to consider that num of addresses is increasing and that we need to have 'fair' policy so that each address is scanned same number of times.

altafan:
>If i understood correctly, we are trying to prevent 'max num of calls exceeded in interval x' which is set on server side(elements).

This doesn't have the purpose of preventing that, but is for sure a needed optimization since the address observations are distributed over time and not executed altogether every X seconds (this causes the max num calls to happen so early in the daemon's lifetime).

>We need to consider that num of addresses is increasing and that we need to have 'fair' policy so that each address is scanned same number of times.

IMO this 'fair' policy is not the right approach... I mean, what's the usefulness of keeping watching every X seconds an address if it results unused for a long while?
In theory, every daemon's address owns only 1 utxo, so the number of addresses to watch is linear with the number of total utxos owned by the wallet (WalletAccount excluded - we don't track utxos for that account). I don't think that stopping watching an address is a smart move of course, but we should definitely consider increasing the interval of an address when it is recognized as UNUSED - like if it doesn't receive funds for Y blocks, for example.

sekulicd:
> This doesn't have the purpose of preventing that ..
My post is concentrated only on elements possible rate limit issue

> IMO this 'fair' policy is not the right approach 
I mentioned this with purpose of having in mind that number of addresses increases and we need to scan them all, and I agree that it would be better if we even enhance it with some kind of 'exponential backoff' for addresses that are not used in longer period, however this would additionally increase complexity of a solution.

tiero:
Would you move this in a dedicated issue please? @sekulicd  I think request throttling is orthogonal to this change. i would keep this for comments for current code.

(Unless you nACK the current PR change and you want to have only the other solution implemented instead of this and not both)

tiero:
> This doesn't have the purpose of preventing that, but is for sure a needed optimization since the address observations are distributed over time and not executed altogether every X seconds (this causes the max num calls to happen so early in the daemon's lifetime).

True, but at the some time we are not 100% sure that distributing ticker in different moment (ie. when user calls the DepositMarket) will prevent multiple ticker overlapping on the same moment.

A worker pool is a sensible approach to make the solution more robust, but IMHO orthogonal to this PR and we can go forward with this PR as low hanging fruit, keeping in mind is not "definitive"

sekulicd:
> Would you move this in a dedicated issue please? @sekulicd I think request throttling is orthogonal to this change. i would keep this for comments for current code.
> 
> (Unless you nACK the current PR change and you want to have only the other solution implemented instead of this and not both)

@tiero I dont have in mind currently solution on how we can control overlapping tickers so that num of calls dont exceed limit, i am still not fully comfortable with new crewler imp (need more time :) ) but if you think it is orthogonal than i am fine with that.

internal/core/application/blockchain_listener.go
```diff
@@ -67,98 +75,172 @@ func newBlockchainListener(
...
 }
 
+func (b *blockchainListener) startPendingObservables() {
+	if len(b.pendingObservables) <= 0 {
+		return
+	}
+
+	for _, observable := range b.pendingObservables {
+		b.crawlerSvc.AddObservable(observable)
+		time.Sleep(200 * time.Millisecond)
```
tiero, line 238: Do you think is 200ms is too much or enought? With Explorer service for sure too much, with ELements too little?

internal/core/application/operator_service.go
```diff
@@ -161,8 +161,8 @@ func (o *operatorService) DepositMarket(
 		numOfAddresses = 1
 	}
 
-	addresses := make([]string, 0, numOfAddresses)
-	blindingKeys := make([][]byte, 0, numOfAddresses)
+	list := make([]AddressAndBlindingKey, numOfAddresses, numOfAddresses)
```
tiero, line 164:
```suggestion
	list := make([]AddressAndBlindingKey, numOfAddresses, len(numOfAddresses))
```

tiero, line 164: why three arguments?

internal/core/application/operator_service.go
```diff
@@ -904,3 +834,66 @@ func (o *operatorService) getAllUnspentsForAccount(
 	}
 	return utxos, nil
 }
+
+func (o *operatorService) observeAddressesForAccount(accountIndex int, list []AddressAndBlindingKey) {
+	for _, l := range list {
+		blindkey, _ := hex.DecodeString(l.BlindingKey)
+		o.blockchainListener.StartObserveAddress(accountIndex, l.Address, blindkey)
+		time.Sleep(200 * time.Millisecond)
```
tiero, line 842: We bring accross this sleep time. We shoudl have it a constant in a single place

internal/core/application/blockchain_listener.go
sekulicd: do we have exact data on limitation from Explorer/Elements?

internal/core/application/operator_service.go
altafan: it's for setting the length of the array already... this way in the loop we don't use append(), but directly assign a value to list[i]

internal/core/application/wallet_service.go
```diff
@@ -169,24 +160,31 @@ func (w *walletService) InitWallet(
...
 			for i, m := range marketsLastDerivedIndex {
-				if err := initVaultAccount(v, domain.MarketAccountStart+i, m, w.crawlerService); err != nil {
+				addresses, err := initVaultAccount(v, domain.MarketAccountStart+i, m)
+				if err != nil {
 					return nil, err
 				}
+				marketAddresses = append(marketAddresses, addresses...)
 			}
 
+			walletAddresses = append(walletAddresses, feeAddresses...)
```
tiero, line 180: why not doing in one single append  both feeAddresses and marketAddresses?

internal/core/application/blockchain_listener.go
altafan: this kind of time.Sleep prevents the observations to be almost concurrent basically... consider that usually the daemon adds more than one address to the watchlist... so these additions are "mitigated" by the sleeping time.

internal/core/application/operator_service.go
altafan: I agree, but since this isn't something we previously agreed on I wanted to know your opinions, guys.

internal/core/application/blockchain_listener.go
tiero: @sekulicd  on the Blockstream.info instance we don't know what they do and they can change unilaterally at they discretion, ban our IPs and so on :P So do not relay on that as info

On the new "Elements as poor man explorer" instead It depends on how you configure the Elements node, but we do not need to expect an optimized node for high availability & high load enviroment.

Those the available conf options

``` 
-rpcthreads=<n>

Set the number of threads to service RPC calls (default: 4) 
       
-rpcworkqueue=<n>

Sets the depth of the work queue to service RPC calls, the default is 16.
```

internal/core/application/wallet_service.go
altafan: the first arg of append is the destination where to add new elements... therefore it's not possible to do something like:
```
walletAddress = append(feeAddresses..., marketAddresses...)
```

https://golang.org/ref/spec#Appending_and_copying_slices

internal/core/application/blockchain_listener.go
tiero: we defenitely know to not "pump" your elements node too much or it will crash :D

https://github.com/bitpay/bitcore-node/issues/463#issuecomment-418575795

altafan: check the [comment right below](https://github.com/bitpay/bitcore-node/issues/463#issuecomment-418592081)... the crash was due to another reason 😅

tiero: Ok cause I found via wasabi https://github.com/zkSNACKs/WalletWasabi/pull/3056

Issue #237, opened 2020-12-02
Fee collection do not work properly


tiero:
At the moment the GetCollectedMarketFee rpc is not calculating the collected fees properly because it's making a sum of the fee basis point per each completed trade for a certain market.

Instead, for each completed trade it should CALCULATE the fee amount.
This is always calculated against the proposer's amount, hence the AmountP of the relative swap request message. Given this, for BUY trades the asset of the fee will be the quote asset of the market, while for those of type SELL the asset will be the base asset of the market.

tiero:
Is it this solved? cannot recall

PR #262, opened 2021-02-26
Fix ReportFee & Drop fee asset

### LINKED ISSUE #241: [Trade] Move updating of market fee in domain scope
At the moment, we're updating the trade domain's MarketFee and MarketFeeAsset at [application level](https://github.com/TDex-network/tdex-daemon/blob/master/internal/core/application/trade_service.go#L374). In order to keep the changes all in one place and not having them sparse around the files, these 2 lines should be moved within the trade.Propose method, at [domain level](https://github.com/TDex-network/tdex-daemon/blob/master/internal/core/domain/trade_service.go#L14), where all changes about a domain should be placed.

The signature of that method should change so that it doesn't expect only the market's quote asset, but also the fee and fee asset (not the entire market because I'm not sure having dependencies between domains is a good choice).

### LINKED ISSUE #242: Drop fee_asset from UpdateMarketFee
The operator cannot decide wich fee asset it should collect the fees, because we can only do on the "way in"﻿

altafan:
This introduces some little changes to fix the ReportMarketFee so that it correctly calculates fees as described by #237.
This also updates the tdex-protobuf dependency, which in its latest version drops the FeeAsset of a market. The same has been done in the daemon, where the FeeAsset field and its related method ChangeFeeAsset have been removed from the domain.Market and accordingly in the rest of the codebase.

Last but not least, the domain.Trade's Propose method has slightly changed with the introduction of another argument in order to set the MarketFee and MarketPrice directly at domain level. Given that, the `Propose()` method calculates the current prices of the swap by simply dividing amountP and amountR and also sets the market fee.

Closes #237.
Closes #241.
Closes #242.

Please @tiero @sekulicd , review this.


internal/core/application/operator_service.go
```diff
@@ -625,19 +620,30 @@ func (o *operatorService) GetCollectedMarketFee(
...
-
-		if val, ok := total[v.MarketFeeAsset]; ok {
-			total[v.MarketFeeAsset] = val + v.MarketFee
+	for _, trade := range trades {
+		feeBasisPoint := trade.MarketFee
+		swapRequest := trade.SwapRequestMessage()
+		feeAsset := swapRequest.GetAssetP()
+		amountP := swapRequest.GetAmountP()
+		_, feeAmount := mathutil.LessFee(amountP, uint64(feeBasisPoint))
+		var marketPrice decimal.Decimal
```
tiero, line 631: we may want to skip the else here, assing the quotePrice by default, if not then assing the base one.

altafan: Done.

tiero (approved):
Agreed with @sekulicd  minor comments, after that ACK to me

PR #267, opened 2021-03-02
Add rescan timestamp && Improve address import

### LINKED ISSUE #265: [Elements explorer] Improve address import/rescan
At the moment, the Elements implementation of the explorer interface makes use of the _importaddress_ RPC to import an address and does not set rescan:true, hence is likely that the daemon won't track old* unspents. Furthermore, this RPC is extremely inefficient in mainnet since scans the entire blockchain for every address.

As an optimization, the explorer should make use of the `importmulti` rpc instead. This RPC makes mandatory to specify the timestamp of the moment since starting to rescan. This turns out to be extremely useful.

The timestamp should be configurable via env with a dedicated variable that defaults the date to dec/01/2020 for now.

(*) old unspents stands for those already owned by an address prior to importing it into the Elements node. Rescan is intended to keep track of them basically.

### LINKED ISSUE #266: [Elements] GetUnspentsForAddresses should pass an array od addresses down to  insetad of calling N times the Elements RPC

altafan:
This contains some improvements to the elements explorer:
 - it now makes use of the `importmulti` RPC to import an address instead of `importaddress`. The benefit is that the former lets one define the height of the blockchain from where the node starts its rescan in search unspents/txs of an imported address, while the latter forces the rescan of the entire blockchain. `importaddress` can take up to 2 hours to complete the rescan for a single address. Such a long time is not admissible of course.
 - In addition, a new env var `TDEX_ELEMENTS_START_RESCAN_TIMESTAMP` can be set to either `0` or any timestamp in Unix format to set the starting point of the rescan.
 - The GetUnspentsForAddresses method does not call N times GetUnspents but instead makes directly use of the `listunspent` RPC.

Closes #265.
Closes #266

Please @tiero review this.


config/config.go
```diff
@@ -134,7 +137,11 @@ func GetNetwork() *network.Network {
 //GetExplorer ...
 func GetExplorer() (explorer.Service, error) {
 	if rpcEndpoint := GetString(ElementsRPCEndpointKey); rpcEndpoint != "" {
-		return elements.NewService(rpcEndpoint)
+		var rescanTime interface{} = "now"
```
tiero, line 140: Isn't better here to have a default if not given?

PR #269, opened 2021-03-03
Optional restore for InitWallet

### LINKED ISSUE #264: [Wallet] Optional restore
At the moment the restoration of the wallet is executed by default. It should become optional by adding a bool field `restore` in the InitWalletRequest of the protobuf that the operator can set to true eventually. Thus, the restore should be skipped by default.

altafan:
Before this, the daemon was always trying to restore the wallet while processing an `InitWallet` request. This has been made optional now. The operator can specify if the wallet must be restore by adding a `restore: true` flag to the request. By default the restoration is skipped.

Closes #264.

Please @tiero review this

### DIFF
--- a/internal/core/application/wallet_service.go
+++ b/internal/core/application/wallet_service.go
@@ -35,6 +35,7 @@ type WalletService interface {
 		ctx context.Context,
 		mnemonic []string,
 		passphrase string,
+		restore bool,
 	) error
 	UnlockWallet(
 		ctx context.Context,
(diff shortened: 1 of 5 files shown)


internal/core/application/wallet_service.go
```diff
@@ -699,10 +701,20 @@ type accountLastDerivedIndex struct {
 }
 
 func (a *accountLastDerivedIndex) total() int {
-	return a.external + a.internal
+	return (a.external + 1) + (a.internal + 1)
```
altafan, line 704: external and internal can take values in the range [-1, N] where -1 means no used addresses and N (variable value) stands for the total num of used addresses found.

Without this fix, I had the daemon panicking for a wallet that had 1 external used address and no internal used. This situation caused the total() method to incorrectly return -1. This fix the problem.

tiero (approved):

PR #270, opened 2021-03-03
Crawler rate limiter

### LINKED ISSUE #263: Implement Request Throttling for explorer service
At the moment we are not sure that our explorer service (either Esplora or Elements) get exhausted and/or rate limit ourself because it can happen many request to come at the some time.

We shoudl prevent that with a mechanism like database's workers pool and so on, that checks the current "request queue" and postpone if a dangerous quota is reached.﻿

Q: How to defined the "dangerous quota"?

A: That's depends on the underlying explorer, I guess 10 observable per second is a good number?

sekulicd:
This limits number of calls per interval, crawler makes to the explorer.

This closes #263 


pkg/explorer/esplora/unspents_test.go
```diff
@@ -44,10 +44,6 @@ func TestGetUnspents(t *testing.T) {
 }
 
 func TestSelectUnspents(t *testing.T) {
```
tiero: It may be a sleep problem maybe?

sekulicd: Not sure how this test is supposed to work, 
if we assume that utxo created after faucet is 1 * math.Pow10(8), how target is targetAmount := uint64(1.5 * math.Pow10(8)) can be bigger? also in assertion we are expecting 2 utxo, maybe I am missing something ...

Also, if i skip this test CI finishes successfully, if not I am getting RACE issue with crawler and this test is passing.
On the other hand locally i dont have issue with crawler RACE condition, only TestSelectUnspents fails with wrong assertions.

pkg/crawler/observable.go
```diff
@@ -18,16 +52,29 @@ func (a *AddressObservable) observe(
...
+	observableStatus *observableStatus,
+	rateLimiter *rate.Limiter,
 ) {
 	if a == nil {
 		return
 	}
 
+	observableStatus.Set(Waiting)
+	err := rateLimiter.Wait(context.Background())
+	if err != nil {
```
altafan, line 64:
```suggestion
	if err := rateLimiter.Wait(context.Background()); err != nil {
```

pkg/crawler/observable.go
```diff
@@ -56,17 +103,28 @@ func (t *TransactionObservable) observe(
 	explorerSvc explorer.Service,
 	errChan chan error,
 	eventChan chan Event,
+	observableStatus *observableStatus,
```
altafan, line 106: do we actually need a Status? why not simplifying with a bool like `isWaiting`.. we're interested in knowing if the current observation is waiting or not.. we seem to don't care at all about its new or processed states.

pkg/crawler/observable.go
```diff
@@ -56,17 +103,28 @@ func (t *TransactionObservable) observe(
...
+	rateLimiter *rate.Limiter,
 ) {
 	if t == nil {
 		return
 	}
 
+	observableStatus.Set(Waiting)
+	err := rateLimiter.Wait(context.Background())
+	if err != nil {
```
altafan, line 115:
```suggestion
	if err := rateLimiter.Wait(context.Background()); err != nil {
```

altafan (changes requested):

pkg/explorer/esplora/unspents_test.go
altafan: they are intended to work because `newTestData()` returns always the same address and key once it has created. So TestGetUnspents should send 1 LBTC to the test address, and then the same should do TestSelectUnspents, so the test should pass because the testing address is owning 2 LBTC... but it doesn't look good as it is to me 😅

pkg/crawler/observable.go
tiero: New would makes sense to eventually drop the same observable if is still waiting (ie. FILO stack)

pkg/crawler/service.go
```diff
@@ -20,26 +22,33 @@ type blockchainCrawler struct {
...
+	ExplorerSvc explorer.Service
+	//crawler interval in milliseconds
+	CrawlerInterval int
+	ErrorHandler    func(err error)
 }
 
 // NewService returns an utxoCrawelr that is ready for watch for blockchain activites. Use Start and Stop methods to manage it.
 func NewService(opts Opts) Service {
+	//10 req per sec
+	rt := rate.Every(100 * time.Millisecond) //TODO config
```
tiero, line 39: make it as parameter of Opts

tiero (changes requested):
Add a config key to throttling and pass down to NewService as Opts parameter

pkg/crawler/observable.go
sekulicd: Done

pkg/crawler/observable.go
sekulicd: Done

pkg/crawler/service.go
sekulicd: Done

sekulicd:
> Add a config key to throttling and pass down to NewService as Opts parameter

Done

config/config.go
```diff
@@ -57,6 +57,12 @@ const (
 	// ElementsRPCEndpointKey is the url for the RPC interface of the Elements
 	// node in the form protocol://user:password@host:port
 	ElementsRPCEndpointKey = "ELEMENTS_RPC_ENDPOINT"
+	// ExplorerLimit represents number of requests per second that crawler
+	//makes to explorer
+	ExplorerLimitKey = "EXPLORER_LIMIT"
+	// ExplorerTokenBurst represents number of bursts tokens permitted from
+	//crawler to explorer
+	ExplorerTokenBurstKey = "EXPLORER_TOKEN_BURST"
```
altafan, line 65: Maybe better to rename `CrawlerLimitKey` and `CrawlerTokenBurst` and relative values accordingly?

tiero: Agreed, but would be consisten with CrawlIntervalKey therefore `CrawlLimitKey` and `CrawlTokenBurst`

sekulicd: Done

tiero:
solve conflicts @sekulicd

sekulicd:
> solve conflicts @sekulicd

Done

tiero (approved):

PR #300, opened 2021-03-12
Macaroon authentication

### LINKED ISSUE #22: gRPC Operator: Macaroon Authentication
although we plan to add an encrypted transport on the trader interface (BOTD#2), we need to secure the operator interface, so remote CLIs or GUIs could be built.

We could implement some gRPC interceptors with very basic user and password (or API Key bearer token) or more advanced solutions like macaroons.
https://pkg.go.dev/github.com/lightningnetwork/lnd/macaroons
https://github.com/lightningnetwork/lnd/blob/master/macaroons/README.md

A bit of history on this choice by lnd to use macaroons https://github.com/lightningnetwork/lnd/issues/20

sekulicd:
This adds macaroon package which is used to authenticate requests.

Commit contains bellow logic:

On startup initialize macaroon service which is in charge of auth.
Macaroon service will be passed to newly introduced auth interceptors which will use it to authenticate requests
Non-operator RPCs are marked as public endpoints
Whoever needs to invoke operator endpoints would need permission files.
It closes #22

NOTE: This is replacement of PR #128 which should be closed.


config/config.go
```diff
@@ -61,6 +63,7 @@ func init() {
 	vip.SetDefault(TradeExpiryTimeKey, 120)
 	vip.SetDefault(DataDirPathKey, defaultDataDir)
 	vip.SetDefault(PriceSlippageKey, 0.05)
+	vip.SetDefault(MacaroonRootKeyPass, "hello")
```
tiero, line 93: I do think we may want to remove default here

config/permissions.go
```diff
@@ -23,10 +23,26 @@ var (
 			Action: "updatemarketstrategy",
 		},
 	}
-	//TODO check permissions
-	readonlyPermissions = []bakery.Op{}
+	ReadonlyPermissions = []bakery.Op{
```
tiero, line 26: ListDepositMarket also?

config/permissions.go
```diff
@@ -107,26 +123,6 @@ var (
 			Entity: "operator",
```
tiero, line 123: Would be possible to not rewrite these again, just inherits like  `...ReadonlyPermissions` ?

cmd/grpc/main.go
```diff
@@ -46,7 +46,7 @@ func main() {
 	}
 	defer macaroonService.Close()
 
-	var pass = []byte("hello") //TODO this should be passed as flag/env?
+	var pass = []byte(config.GetString(config.MacaroonRootKeyPass))
```
tiero: I would move this in the Unlock RPC, so we can re-use that wallet password also for macaroons. Doable?

tiero (changes requested):
Merging this will irremediably break the `tdex` operator CLI. 

We must put effort in update all the clients, this also serves as documentation.

config/permissions.go
```diff
@@ -0,0 +1,214 @@
...
+			Entity: "operator",
+			Action: "listmarket",
+		},
+		{
+			Entity: "operator",
+			Action: "balancefeeaccount",
+		},
+	}
+
+	AdminPermissions = []bakery.Op{
```
altafan, line 45: Admin is missing permissions for `claimfeedeposit` and `claimmarketdeposit`. Your branch needs to be synced with master.

config/permissions.go
```diff
@@ -0,0 +1,214 @@
...
+		{
+			Entity: "operator",
+			Action: "reportmarketfee",
+		},
+	}
+)
+
+// RPCServerPermissions returns a mapping of the main RPC server calls to
+// the permissions they require.
+func RPCServerPermissions() map[string][]bakery.Op {
```
altafan, line 131: shouldn't it be?
```suggestion
func RPCServerPermissions() map[string]bakery.Op {
```

I didn't get why mapping each RPC path to a slice of permissions.

internal/interfaces/grpc/interceptor/interceptor.go
```diff
@@ -2,23 +2,33 @@ package interceptor
...
 	dbbadger "github.com/tdex-network/tdex-daemon/internal/infrastructure/storage/db/badger"
+	"github.com/tdex-network/tdex-daemon/pkg/macaroons"
 	"google.golang.org/grpc"
 )
 
 // UnaryInterceptor returns the unary interceptor
-func UnaryInterceptor(dbManager *dbbadger.DbManager) grpc.ServerOption {
+func UnaryInterceptor(
+	dbManager *dbbadger.DbManager,
+	macaroonService macaroons.Service,
```
altafan, line 14: it would be better to pass the pointer to the macaroon service so there's no need to dereference it elsewhere.

pkg/macaroons/constraints.go
```diff
@@ -0,0 +1,94 @@
...
+	return newMac, nil
+}
+
+// Each *Constraint function is a functional option, which takes a pointer
+// to the macaroon and adds another restriction to it. For each *Constraint,
+// the corresponding *Checker is provided if not provided by default.
+
+// TimeoutConstraint restricts the lifetime of the macaroon
+// to the amount of seconds given.
+func TimeoutConstraint(seconds int64) func(*macaroon.Macaroon) error {
```
altafan, line 46:
```suggestion
func TimeoutConstraint(seconds int64) Constraint {
```

pkg/macaroons/constraints.go
```diff
@@ -0,0 +1,94 @@
...
+		requestTimeout := time.Now().Add(time.Second * macaroonTimeout)
+		caveat := checkers.TimeBeforeCaveat(requestTimeout)
+		return mac.AddFirstPartyCaveat([]byte(caveat.Condition))
+	}
+}
+
+// IPLockConstraint locks macaroon to a specific IP address.
+// If address is an empty string, this constraint does nothing to
+// accommodate default value's desired behavior.
+func IPLockConstraint(ipAddr string) func(*macaroon.Macaroon) error {
```
altafan, line 58: sam here

pkg/macaroons/constraints.go
```diff
@@ -0,0 +1,94 @@
...
+				macaroonIPAddr.String())
+			return mac.AddFirstPartyCaveat([]byte(caveat))
+		}
+		return nil
+	}
+}
+
+// IPLockChecker accepts client IP from the validation context and compares it
+// with IP locked in the macaroon. It is of the `Checker` type.
+func IPLockChecker() (string, checkers.Func) {
```
altafan, line 75:
```suggestion
func IPLockChecker() Checker {
```

altafan (changes requested):
`config` **must not** be imported within `pkg`. ATM, on the contrary, _macaroons_ imports it here in the implementation of the Init() and genMacaroons functions.

config/config.go
sekulicd: Done

tiero:
closing in favor of #370

Issue #279, opened 2021-03-04
[Badger] Move unspents to in-memory mode


tiero:
Benefits:

- Lower disk space
- Arguably faster reads and somewhat comparable write times to BadgerDB
- Faster trades and faster Balances RPC response (It could take 14s sometimes!!!)

Cons:

- Increase in RAM/CPU(?) usage
﻿- Slower startup time, but una tantum (recreate the inmemory repo from explorer) 

To do this, we need to reload utxos #299 at daemon startup to recreate the utxo set in-memory (after shutdown or crash)

altafan:
> Increase in RAM/CPU(?) usage
>  ﻿- Slower startup time, but una tantum (recreate the inmemory repo from explorer)

Not sure about this.. What I mean is that at this point we might be interested in storing ONLY the actual unspent whether locked or unlocked.  This would speed up also the restoration: since we already have all the used addresses in storage, it should just be enough to retrieve their unspents, and if done with Esplora in particular, it would return only actual unspent utxos.

Given this, when an unspent is spent we could physically remove it from the storage. At the moment the deletion is logical since we mark those unspents as Spent. This would lead to a smaller utxo set that takes less space on RAM.

There are pros and cons for both logical and physical deletion; this is a nice [SO](https://stackoverflow.com/questions/378331/physical-vs-logical-hard-vs-soft-delete-of-database-record) in which the various answers point out different valid aspects, but it seems to me that in our case the main concern is the impact on memory, so physical deletion might be the preferred choice.

PR #301, opened 2021-03-12
Refactor domains and repos

altafan:
This adds changes to the domain package in order to not be dependent directly on config and other dependencies (useful also when testing to be able to mock them).
For this reason, the following interfaces and relative variables have been added:
 * IMnemonicStore - MnemonicStore for the plaintext mnemonic
 * IEncrypter - Encrypted for the encryption/decryption
 * ISwapParser - SwapParser to serialize/deserialize swap messages to/from binary representation (raw bytes)
 * IPsetManager - PsetManager to extract info like txid or txhex from swaps' transactions
 * SwapRequest, SwapAccept, SwapComplete to decouple swap representations from grpc stubs

These changes, as already introduced, make it possible to create simpler mocks than those already in place for tests, and, furthermore, to extend the testing coverage, in particular for failing situations.

For repositories tests, there is now no need to test any but the "happy path" (ie. only the non-failing cases) since errors can be raised only at domain level and forwarded by them just in case.

The only exception is relative to the DbManager, which could raise an error in the event that the Transaction fails to commit for any reason. There are already tests in place to check these cases, but they have been commented at the moment.
The reason is that the in-memory DbManager does not support transactional read/write operation at the moment, therefore I preferred to skip these tests for both instead of adding weird logic, it seemed cleaner to me.  I marked them with TODOs, fixing in-memory transactional is out of scope here.

Feel free to checkout this locally and uncomment them to see that they work at least for badger.

TODO: application tests. I thought it was better to at least open the PR so that this part of the task can be validated. Meanwhile, I'm going to refactor application tests in order to have a component test checking that at application level it could be possible to reproduce the complete flow to start the daemon, open a market and make concurrent trades. I think this is just enough. Our actual "integration" test is something in between a component test (that is a sort of e2e test at application level, in the sense that it shouldn't test narrow behavior of a service, but rather the flow(s) in which it's involved) and an e2e test.
There's no evident benefit in testing such flows through a specific port (gRPC), and on the other hand, we are forced to make use of a grpc client to make it real. This increases the overhead of this approach has more downsides than advantages.
There's no interest in testing ports at all I think. They are just translator from low-level to high-level types so they could leave untested IMO.

Please @tiero @sekulicd @louisinger please review this.


internal/core/application/blockchain_listener.go
```diff
@@ -391,9 +391,10 @@ func (b *blockchainListener) checkMarketAccountFundings(ctx context.Context, acc
 			}
 			log.Warnf("%s asset is missing for market %d", asset, accountIndex)
 		case 2:
+			baseAsset := config.GetString(config.BaseAssetKey)
```
tiero, line 394: at shi point, wouldn't be better to be a filed of the `blockchainListener` itself? So we can entirely decouple from config

internal/core/application/errors.go
```diff
@@ -9,4 +9,6 @@ var (
 	)
 	// ErrUnknownStrategy ...
 	ErrUnknownStrategy = errors.New("strategy not supported")
+	// ErrMarketNotExist ...
+	ErrMarketNotExist = errors.New("market does not exists")
```
tiero, line 13:
```suggestion
	ErrMarketNotExist = errors.New("market does not exist")
```

internal/core/application/operator_service.go
```diff
@@ -148,15 +148,19 @@ func (o *operatorService) DepositMarket(
 		}
 
 		nextAccountIndex := latestAccountIndex + 1
-		if _, err := o.marketRepository.GetOrCreateMarket(ctx, nextAccountIndex); err != nil {
+		fee := int64(config.GetFloat(config.DefaultFeeKey) * 100)
```
tiero, line 151: Also this should become a field of `operatorService` and we pass down when we create it

internal/core/application/test_util.go
```diff
@@ -67,7 +67,7 @@ func newMockServices(
 	// create a market repo
 	marketRepo := inmemory.NewMarketRepositoryImpl(dbManager)
 	if !marketRepositoryIsEmpty {
-		err := fillMarketRepo(ctx, &marketRepo, initPluggableMarket)
+		err := fillMarketRepo(ctx, marketRepo, initPluggableMarket)
```
tiero, line 70: are we sure we need to pass the pointer to it?

internal/core/domain/trade_model.go
```diff
@@ -1,64 +1,299 @@
...
+	GetFailureMessage() string
+}
+
+// Swap is the data structure that represents any of the above swaps.
 type Swap struct {
-	ID      string
-	Message []byte
+	ID        string
+	Message   []byte
+	Timestamp uint64
```
tiero, line 282: what this timestamp refers to?

internal/core/domain/trade_model.go
```diff
@@ -69,14 +304,15 @@ type Trade struct {
 	PsetBase64       string
 	TxID             string
 	TxHex            string
-	Timestamp        Timestamp
+	ExpiryTime       uint64
+	SettleTime       uint64
```
tiero, line 308:
```suggestion
	SettlementTime       uint64
```

internal/core/domain/trade_service.go
```diff
@@ -4,76 +4,80 @@ import (
...
 func (t *Trade) Accept(
 	psetBase64 string,
 	inputBlindingKeys,
 	outputBlindingKeys map[string][]byte,
 ) (bool, error) {
+	if t.Status.Code >= Accepted {
+		return true, nil
+	}
+
+	if t.IsExpired() {
```
tiero, line 60: Not sure about this: the exiration is "attached" after the swap is accepted by the daemon.

tiero: A SwapRequest from a client has no expiration: as long the pricing terms are good, we move forward

internal/core/domain/trade_service.go
```diff
@@ -82,226 +86,187 @@ func (t *Trade) Accept(
...
-	}); err != nil {
-		t.Fail(
-			t.SwapAccept.ID,
-			FailedToCompleteStatus,
-			pkgswap.ErrCodeFailedToComplete,
-			err.Error(),
-		)
-		return &CompleteResult{OK: false}, nil
+	if t.IsExpired() {
+		if !t.Status.Failed {
```
tiero, line 113: why we would prevent to write again Failed?

internal/core/domain/trade_service.go
```diff
@@ -82,226 +86,187 @@ func (t *Trade) Accept(
...
 	t.TxHex = txHex
 	return &CompleteResult{OK: true, TxHex: txHex, TxID: txHash}, nil
 }
 
-func (t *Trade) Settle(settlementTime uint64) error {
-	t.Status = CompletedStatus
-	return t.AddBlocktime(settlementTime)
-}
+// Settle brings the trade from the Completed to the Settled status, unsets the
+// expiration time and adds the timestamp of the settlement (it could be
```
tiero, line 147:
```suggestion
// expiration time and adds the timestamp of the settlement (it must be
```

internal/core/domain/trade_service.go
```diff
@@ -82,226 +86,187 @@ func (t *Trade) Accept(
...
-func (t *Trade) Fail(swapID string, tradeStatus Status, errCode pkgswap.ErrCode, errMsg string) {
-	swapFailID, swapFailMsg, _ := pkgswap.Fail(pkgswap.FailOpts{
-		MessageID:  swapID,
-		ErrCode:    errCode,
-		ErrMessage: errMsg,
-	})
-	t.SwapFail.ID = swapFailID
-	t.SwapFail.Message = swapFailMsg
-	t.Status = tradeStatus
+	if t.IsExpired() {
```
tiero, line 154: Here is tricky: we consider a swap completed (ie. with all the signatures) expired if for some reasons to be included in a block it takes more time, not dependent on either the user and the daemon?

tiero:
I do think we should enforce test for badger and just skip the ones for the in-memory repos altogether

internal/core/domain/trade_model.go
altafan: before this, Timestamp was a struct for those of the swaps and the expiration... this moves the swap timestamps inside the Swap struct and 2 ad hoc fields for the expiration and the settlement.

internal/core/domain/trade_service.go
altafan: this just make sure that the trade is set to fail status if expired. The check just prevent setting the status to Failed if it already is.

altafan:
> I do think we should enforce test for badger and just skip the ones for the in-memory repos altogether

Not sure about that. Why should we skip testing in-memory repos if we planned to use them (#279)?

tiero:
> > I do think we should enforce test for badger and just skip the ones for the in-memory repos altogether
> 
> Not sure about that. Why should we skip testing in-memory repos if we planned to use them (#279)?

My idea was to use still using badger as infra, just switching the unspent to in-memory mode? Unless we do not need "fancy" querying/tx capabilities.

internal/core/domain/trade_model.go
```diff
@@ -1,64 +1,299 @@
...
+	Code int
+}
+
+func (s *SwapError) Error() string {
+	return s.Err.Error()
+}
+
+// ISwapParser defines the required methods to override the default swap
+// message parser, which is grpc-proto.
+type ISwapParser interface {
```
sekulicd, line 80: I  dont see benefit of grouping these methods as interface:
1. we wont have more implementations of it
2. it doesnt have impact on testing

also i would methods to domain/trade_service.go

internal/core/domain/errors.go
```diff
@@ -1,56 +1,91 @@
...
+	// ErrVaultAlreadyInitialized ...
+	ErrVaultAlreadyInitialized = errors.New("vault is already initialized")
+	// ErrVaultNullMnemonicOrPassphrase ...
+	ErrVaultNullMnemonicOrPassphrase = errors.New("mnemonic and/or passphrase must not be null")
+	// ErrVaultNullNetwork ...
+	ErrVaultNullNetwork = errors.New("network must not be null")
+	// ErrVaultAccountNotFound ...
+	ErrVaultAccountNotFound = errors.New("account not found")
+)
+
```
sekulicd, line 67: i guess we could add comment (// Trade errors) here to be aligned with L50

sekulicd: also bellow where we grouped errors in separate var block based on domain

internal/core/domain/vault_model.go
```diff
@@ -37,30 +38,98 @@ type AddressInfo struct {
...
+	return len(config.GetMnemonic()) > 0
+}
+
+func (c configStore) Get() []string {
+	return config.GetMnemonic()
+}
+
+// IEncrypter defines the required methods to override the default encryption
+// performed through pkg/wallet package.
+type IEncrypter interface {
```
sekulicd, line 75: I tried to 'google' bit about adding Prefix 'I' for interfaces, there are different views, I personally dont like it and in this specific place I would use only Encrypter, but there are cases when there is sturct and intercafe with same name and then using 'I' prefix helps, but even in that way I would try  to find some better name.
However dont have clear opinion on this.

internal/core/domain/vault_model.go
```diff
@@ -37,30 +38,98 @@ type AddressInfo struct {
...
+	Encrypt(mnemonic, passphrase string) (string, error)
+	Decrypt(encryptedMnemonic, passphrase string) (string, error)
+}
+
+type walletEncrypter struct{}
+
+func (w walletEncrypter) Encrypt(mnemonic, passphrase string) (string, error) {
```
sekulicd, line 82: here we are just wrapping wallet Encrypt and Decrypt methods and creating interfaces, i guess goal was to mock Encrypt and Decrypt methods in tests, i think exposing of an interface was supposed to be done in wallet package itself

tiero:
About new "style" changes introduced int his PR (ie. "I" for interfaces, "With" instead of "By"): @altafan 

a) always stick with the style you go already in master, address only functional changes
b) if you don't like for any reasons, opens up a dedicated PR to address only the style changes

I understand is a bummer, but please can you rollback to the stylistic changes? no time for holy wars :)

internal/core/domain/vault_model.go
tiero: Not saying one way is better that other, but style changes should be done in a dedicated PR, not mixed with functional/testing stuff

internal/core/domain/vault_model.go
tiero: Not saying one way is better that other, but style changes should be done in a dedicated PR, not mixed with functional/testing stuff.

internal/core/domain/trade_model.go
altafan: it does have impact on testing instead. Without abstracting components like the swap parser it wouldn't just be possible to use mocks when testing. This is the reason why we had "real" mocks for trade domain tests before this. Take a look at them before and after this PR. I do think these abstractions bring benefits, actually, in particular when testing.

Also, this is a nice way to be able to select which type of MnemonicStore we want to use. At the moment the default MnemonicStore is the "config" one, which lives inside the Vault domain. They should be decoupled from each other, though, so the plan would be to move that implementation outside the domain. This wouldn't just be possible without such abstractions. But, guys, if you have another solution in mind, please share so we can discuss about that.

internal/core/domain/vault_model.go
altafan: Encrypt and Decrypt are static functions of the wallet package. I think the encrypter interafce is more related to the domain rather than the wallet package itself. After all, it's required by the Vault domain and it's there that we need to define how the encrypter should look like from a domain-level point of view.

internal/core/domain/trade_service.go
altafan: hmmm.. actually, the expiration time was set for the trade within the Proposal method even before this PR.
I'm going to fix this then.

internal/core/domain/trade_service.go
altafan: yeah, you're right.. if for some reason a completed trade exceeds the expiration time to be included in the blockchain it couldn't be marked as settled by the daemon. Shall we just remove this check then?

altafan: Question: should we enforce that the provided settlemetTime (de facto a blocktime) is actually greater than the complete time?

internal/core/domain/mocks_test.go
```diff
@@ -12,86 +11,228 @@ import (
...
+func (m mockPsetParser) GetTxHex(psetBase64 string) (string, error) {
+	args := m.Called(psetBase64)
+	var res string
+	if a := args.Get(0); a != nil {
+		res = a.(string)
+	}
+	return res, args.Error(1)
 }
 
+// /*
```
tiero, line 152: Remove commented code

internal/core/domain/trade_service.go
tiero: I would not enforce just because is not something that should block the trade: if it hapens something nasty with wither our machine clock or with the explorer block time, so lets favor trading instead of logical correctness

tiero:
ACK current domain refactor!

Let's move ASAP on component tests

internal/core/domain/trade_service.go
altafan: I agree indeed. Already pushed the changes to check the expiration only in Complete() where the trade must be brought from Accepted to Completed before a certain expiration timestamp.

Issue #298, opened 2021-03-10
Update "spentness" after RPC calls to ClaimMarketDeposit, ClaimFeeDeposit, WithdrawMarket & TradeComplete


tiero:
With #296 and #297 in place we need to define moments that will update the Unspent repository with either newly found unspent and to mark them as spent﻿

tiero:
Actually I forgot that there is another case where "spentness" could be changed: The trader is not forced to call the `TradeComplete`, he could by himself broadcast the transaction right away. 

We [should still use the crawler to go fetch a txid after TradePropose ](https://github.com/TDex-network/tdex-daemon/blob/master/internal/core/application/trade_service.go#L401) and therefore we could have the crawler to still update the unspents repository.

My idea is to not watch continuously, but just go watch one time only after the [ExpiryTime](https://github.com/TDex-network/tdex-daemon/blob/master/internal/core/application/trade_service.go#L380): If the tx is NOT in mempool by then, we can leave it to the TTL to "unlock" the unspents, otherwise we go to mark as spent that linked utxos.

altafan:
> My idea is to not watch continuously, but just go watch one time only after the ExpiryTime: If the tx is NOT in mempool by then, we can leave it to the TTL to "unlock" the unspents, otherwise we go to mark as spent that linked utxos.

what is supposed to be this "one time after the ExpiryTime"? What if in this time frame the trade service receives a proposal request?
In this way, we might risk making the daemon to try double-spending utxos that are actually spent while they would be recognized as unspent simply because the bc listener hasn't done its job yet.

tiero:
a) Trade `A` gets accepted at `t0`
b) it locks utxo `X` taken from the utxo set `U` with expiry time `t5`
c) a ticker is attached to trade `A` txid's to be fired a `t5+` to go check on chain if it exists in mempool
d) new utxo set becomes  `N` which is `U - X`
e) Trade `B` comes in and gets accepted at `t1`
f) it locks utxo `Y` from utxo set `N`  with expiry time `t6`
g) a ticker is attached to trade `B` txid's to be fired a `t6+` to go check on chain if it exists in mempool
h) new utxo set becomes  `M` which is `N - Y`
h) we are `t6` exiprty time for trade `A`, the ticker fire up and goes on chain and finds nothing: we can unlock the utxo `X` locked by trade `A` so we have utxo set `L` which is `M + X`


[...]

PR #302, opened 2021-03-13
Claim Fee/Market deposit RPC

### LINKED ISSUE #296: Add ClaimMarketDeposit  to create a new market  and ClaimFeeDeposit RPCswith to fund the fee account
## proto spec

```proto
service Operator {
	// Operator can provide transaction(s) outpoints of deposits made to create a new market.
	// The transaction must be visible and confirmed, otherwise an error will be returned,
	// inviting the operator to retry again
	rpc ClaimMarketDeposit(ClaimMarketDepositRequest) returns (ClaimMarketDepositReply)
	rp ClaimFeeDeposit(ClaimFeeDepositRequest) returns (ClaimFeeDepositReply)
}

message ClaimMarketRequest {
	Market market = 1; // this may or may not exists already. Use DropMarket to delete an unclaimed market
	repeated TxOutpoint outpoints = 2; //list of tx outpoints with deposit outputs. If many vout in the same transaction, the tx hash will be repeated
}

message TxOutpoint {
	string tx_hash = 1;
	int32 index = 2;	
}

message ClaimMarketReply {}

message ClaimFeeDepositRequest {
        repeated TxOutpoint outpoints = 1; //list of tx outpoints with deposit outputs. If many vout in the same transaction, the tx hash will be repeated
}
message ClaimFeeDepositReply {}
```

## Logic

These RPCs will trigger the daemon to go check those provided transactions and to recognize if the given outpoint are locked on addresses/scripts owned by the daemon. If this validation goes through, it will not return error and will in async way go update the spentness of the utxo set #298 decoding the transactions hexes

sekulicd:
This adds implementation for ClaimMarketDeposit and ClaimFeeDeposit RPCs.

This closes #296 

@tiero, @altafan  please review.


tiero:
Yes the Market  internal entry MUST always be created via DepositMarket indeed, its how an operator get to know of addrresses for his deposists.

In the ClaimDepositMarket, the addresses/script of the given outpoint become the way we link those txs to rhe internal market representation. Ie. accountIndex

It looks like we may need to query the Vault with a "GetAccountIndexWithDerivedAddresses"

tiero:
> Funding Market

That's become the job of the ClaimDepositMarket indeed: got the tx outpoints from.the operators, it must verify if valid, then fund the market

tiero:
> Checking account balance

 Its the job of #298 after a "write" that likely updates our balances, we go check also for the fee account balance treshold and eventually log

internal/core/application/operator_service.go
```diff
@@ -826,75 +827,147 @@ func (o *operatorService) FeeAccountBalance(ctx context.Context) (
...
+			return domain.ErrInvalidBaseAsset
+		}
 
-	addresses, bkPairs, err := o.vaultRepository.
-		GetAllDerivedAddressesAndBlindingKeysForAccount(
+		if err := validateAssetString(market.QuoteAsset); err != nil {
+			return domain.ErrInvalidQuoteAsset
+		}
+
+		if market.BaseAsset != config.GetString(config.BaseAssetKey) {
```
tiero, line 857: Just a notice: In the #301 we are decoupling config package from app services, will leave it for now, but will eventually need to be rebased after we do the change in that PR

internal/core/application/operator_service.go
```diff
@@ -814,6 +827,293 @@ func (o *operatorService) FeeAccountBalance(ctx context.Context) (
...
+				)
+			if err != nil {
+				return err
+			}
+			addressesPerAccount[accountIndex] = addresses
+			bkPairs = bk
+		} else {
+			return domain.ErrMarketNotExist
+		}
+	} else {
```
tiero, line 882: What  happens if the given market's asset hashes are empty? I do think we shoudl just return error to user. He will need to call the DepsitMarket first to create a new entry and fund it.

sekulicd: if asset hashes are emty this means that Market is still not funded but DepositMarket is called

internal/core/application/operator_service.go
```diff
@@ -814,6 +827,283 @@ func (o *operatorService) FeeAccountBalance(ctx context.Context) (
...
+	if marketReq.BaseAsset != config.GetString(config.BaseAssetKey) {
+		return domain.ErrInvalidBaseAsset
+	}
+
+	market, accountIndex, err := o.marketRepository.GetMarketByAsset(ctx,
+		marketReq.QuoteAsset)
+	if err != nil {
+		return err
+	}
+
```
tiero, line 864: Here it shoudl be commented what you are trying to do (ie. group addresses per accountIndex) and why you are doing

internal/core/application/operator_service.go
```diff
@@ -814,6 +827,283 @@ func (o *operatorService) FeeAccountBalance(ctx context.Context) (
...
+	} else {
+		vault, err := o.vaultRepository.GetOrCreateVault(
+			ctx,
+			nil,
+			"",
+		)
+		if err != nil {
+			return err
+		}
+		allDerivedAddressInfo := vault.AllDerivedAddressesInfo()
```
tiero, line 885: Would the Vault repository have this method that returns all dervided addresses, already grouped by account index?

internal/core/application/operator_service.go
sekulicd: Done, exposed in new address which name is self explanatory

internal/core/application/operator_service.go
sekulicd: Explained offline

internal/core/application/operator_service.go
```diff
@@ -814,6 +827,294 @@ func (o *operatorService) FeeAccountBalance(ctx context.Context) (
...
+		if err != nil {
+			return err
+		}
+		addressesBlindingKeys := vault.AddressesBlindingKeysGroupByAccount()
+		for k, v := range addressesBlindingKeys {
+			addressesPerAccount[k] = v.Addresses
+			bkPairs = v.BlindingKeys
+
+		}
+	}
```
altafan, line 886: I don't think this is the best approach... I mean, our goal is to discover what's the account index of the not-yet-funded market that is supposed to own the provided outpoints.

Retrieving ALL the addresses from the vault domain means including also those of the fee, wallet, and already-funded markets accounts. And what's more, for every account also internal addresses. But deposits can be sent only to those external!! All these things slow down this discovery process (too much maybe).

We are still searching for a non-funded market so there's must be a better way.

I propose, instead, to get the list of all markets (marketRepo.GetAllMarkets()), filter it to focus only on those not yet funded, and only for these accounts asking the vault domain for their derived addresses (only those external!!!). What do you think guys?

internal/core/application/operator_service.go
```diff
@@ -814,6 +827,294 @@ func (o *operatorService) FeeAccountBalance(ctx context.Context) (
...
+var (
+	marketDeposit depositType = "MarketDeposit"
+	feeDeposit    depositType = "FeeDeposit"
+)
+
+// ClaimMarketDeposit method add unspents to the market
+func (o *operatorService) ClaimMarketDeposit(
+	ctx context.Context,
+	marketReq Market,
+	outpoints []TxOutpoint,
```
altafan, line 841: if we used the domain.OutpointWithAsset type here, the marketReq argument would become useless.

I think this would make the life of the operator a bit easier.. he would be only concerned with adding reference to some utxos and their asset without adding a market or leaving it empty.

internal/core/application/operator_service.go
tiero: Good suggestion to filter only for non-funded account, but I do think could still be a method of the vault repo

internal/core/application/operator_service.go
tiero: I do believe the opposite, is better to have the operator to be explicit in what market is going to claim and prepares him to all future workflows will eventually require already him to specify the market always., it works as self-documenting and self-confirming action.  I need this for operator frontends, gives a better clear framework for making dashabord etc..

internal/core/application/operator_service.go
altafan: how would it be possible? I mean, the market repo is the one storing domain with which is possible to know if a market is funded or not. How could it be possible to retrieve this info within the Vault repository?

tiero: a) we get account indexes of non funded markets via market repo (ie. the one created by DepositMakert but with empty base/quote asset)
b) we ask to the vault domain to retrieve us only the external addresses of those account indexes
c) we compare to the outpoints given in the ClaimDespositMarket RPC request

tiero: so we need both Market and Vault repo ofc, the point is to have a method on the Vault repo that group by account the external addresses retrieved, since we need later

tiero:
Resolve conflicts @sekulicd

internal/core/application/operator_service.go

internal/core/domain/vault_service.go
```diff
@@ -151,10 +151,23 @@ type AddressesBlindingKeys struct {
 
 // AddressesBlindingKeysGroupByAccount returns all address, blinding key pairs
 //grouped by account
-func (v *Vault) AddressesBlindingKeysGroupByAccount() map[int]AddressesBlindingKeys {
+func (v *Vault) AddressesBlindingKeysGroupByAccount(
+	accountIDs []int,
```
tiero, line 155:
```suggestion
	accountIndexes []int,
```

internal/infrastructure/storage/db/inmemory/market_memory_repository_impl.go
```diff
@@ -65,6 +65,22 @@ func (r MarketRepositoryImpl) GetTradableMarkets(_ context.Context) (tradableMar
 	return tradableMarkets, nil
 }
 
+//GetNonTradableMarkets returns all the markets not  available for trading
+func (r MarketRepositoryImpl) GetNonTradableMarkets(
```
tiero, line 69: Well, here we need to get the non funded market, rather non tradable: a market could be funded but closed bacause operator decided so

internal/core/domain/vault_service.go
sekulicd: Done

internal/infrastructure/storage/db/inmemory/market_memory_repository_impl.go
sekulicd: Done

tiero (approved):

Issue #303, opened 2021-03-16
TradeComplete: respond ASAP to clients and update Trades repo in async


tiero:
At the moment we use a "consistent" but very slow solution to implement the TradeComplete, we use a slow [Read-write transaction](https://dgraph.io/docs/badger/get-started/#read-write-transactions) to process a SwapComplete message, when the only thing we need is to do a [Read-only transaction](https://dgraph.io/docs/badger/get-started/#read-only-transactions) to check the given SwapComplete contains the right countersigned transaction. (ie. GetTradeBySwapAcceptID)

We should broadcast the transaction right away and respond ASAP to the user once we got the response from explorer, without waiting for the [UpdateTrade](https://github.com/TDex-network/tdex-daemon/blob/master/internal/core/application/trade_service.go#L420) to complete his database write. In this regard should be fired as goroutine non-blocking and we should handle ourself possible tx conflicts and retry, this should NEVER compromise the trading flow and the user, being thrown up to them interan errors.

tiero:
Re: tx conflicts on prod daemon

I do think is the UpdateMarketPrice is contending the lock on the Store with the TradeComlete, it is trying to do a read-write transaction just to get the account index to reference it in the Prices store. This provably reason we do not spot on local testing.

As a nice enhancement, It looks like to me we should split Trade in his own db store (like prices, unspents) so the Market could be read by other RPCs without causing lock contentions.

PR #304, opened 2021-03-16
Make trade write non blocking in TradeComplete

tiero:
Not tested yet, added this commit to show with code what I meant in  with https://github.com/TDex-network/tdex-daemon/issues/303

AFAIK UpdateTrade inherithing the contect from interceptor, will eventually retry. an addition could be to handling the error in the goroutine to trigger infinite retries until the db release the lock

Please @altafan @sekulicd review﻿


sekulicd:
By watching the code it looks to me that UpdateTrade wont retry if tx conflict happens, cause handler wont return error based on which tx is retried in [here](https://github.com/tiero/tdex-daemon/blob/80b88aeae351d95c0349969012fd0fc5bd35960f/internal/infrastructure/storage/db/badger/db_manager.go#L135)

tiero:
Now it should be able to retry the write in case of conflicts. This woudl eventually change if we agreed to work on #305

tiero:
Taking into cosideration transacion retry, in this case of `persistTradeCompleted` where I do not need to throw up any conflict error to the client (not a user concern whether or not our internal Trade has been persisted or not) we can can say we do not need to care this at the app layer at all, but delegate to `tradeRepository.UpdateTrade` to "manage" it for us.

So in this case would be better maybe to assumen that if  `UpdateTrade` does not have any "tx" field in the `Context` is an implicit way to "hey please repo, don't care about tx conflicts, please retry on my behalf"

This is true for many other occasions where we need "managed" transaction ie. we do not source it from the grpc handler

cc @sekulicd @altafan

internal/infrastructure/storage/db/badger/trade_repository_impl.go
```diff
@@ -109,7 +111,15 @@ func (t tradeRepositoryImpl) UpdateTrade(
 		return err
 	}
 
-	return t.updateTrade(ctx, updatedTrade.ID, *updatedTrade)
+	for {
+		err := t.updateTrade(ctx, updatedTrade.ID, *updatedTrade)
+		if err != nil {
+			if isManaged && isTransactionConflict(err) {
```
sekulicd, line 117: based on above comment:
`if !isManaged && isTransactionConflict(err) {`

internal/infrastructure/storage/db/badger/trade_repository_impl.go
```diff
@@ -99,6 +99,8 @@ func (t tradeRepositoryImpl) UpdateTrade(
 	ID *uuid.UUID,
 	updateFn func(t *domain.Trade) (*domain.Trade, error),
 ) error {
+	isManaged := ctx.Value("tx") == nil
```
sekulicd, line 102: isManaged -> means someone outside manages tx and that it should be:
`	isManaged := ctx.Value("tx") != nil
`

tiero: ok I have a different "semantic" in mind :) I mean managed from an external point of view: the repository method "manages" the creation, commit, rollback and retry for you, without caring to give us.

Maybe better calling `txIsGiven`

tiero: done

tiero:
Removed transaction from grpc handler as requested offline by @sekulicd  since we "manage" it in the repository

internal/core/application/trade_service.go
```diff
@@ -445,40 +445,47 @@ func (t *tradeService) tradeComplete(
...
-			txID = res.TxID
+	if _, err = t.explorerSvc.BroadcastTransaction(res.TxHex); err != nil {
+		return
+	}
 
-			log.Info("trade with id ", tradeID, " broadcasted: ", txID)
+	txID = res.TxID
+	log.Info("trade with id ", tradeID, " broadcasted: ", txID)
+
+	go t.persistTradeCompleted(trade)
```
sekulicd, line 472: we just updated trade here, maybe it would be cleaner if we handle this in anonymous func, than having new method called persistTradeCompleted which is actually only updates trade.

tiero: done

sekulicd (commented):
ACK

altafan:
nACK.

The purpose of db txs is to make several changes to different repositories in a transactional way.
Let's take for example TradeService.TradePropose() of the app layer. It involves changes to different stores, like the Vaut, the Trade and even the Unspent repositories (not yet but soon).
If for any reason a change to any of those repos causes a conflict, in order to keep the storage consistent any other change to other repo must be rolled back in accordance.

This is why for historical reasons you can find the DbManager at interface (presentation) layer: we thought it was better to retry invoking the same app method again to reproduce the same changes in case of conflicts basically.

About the changes introduced by this:
* we could say that `UpdateTrade` stands to `TradeRepository` as `RunTransaction` stands to `DbManager` in some way: what I mean is that `UpdateTrade` is intended to apply several changes to the same domain in a transactional way, and in case any of the changes fails, the trade is rolled back to its previous state, just like a normal db tx does with commit/rollback machinery.
Said this, although there are no actual errors on the changes introduced in `TradeService.tradeComplete()`, it's a conceptual error to make use of _GetOrCreateTrade_, make changes outside _UpdateTrade_ and simply return the modified trade in it.
As also it is an error to check for tx conflicts within TradeBadgerRepository.UpdateTrade(), while the tx is actually committed somewhere else. There, we basically check if the tx is conflicting before it's actually committed.


It seems to me that, especially for badger, we are moving to a situation where we have one store (thus exactly one file) per each repository. While this seems the best way to avoid tx conflicts, on the other side increases the complexity of making cross-repo write operations in a transactional way.

We decided to split badger stores because in the past we had different actors performing different write operations on repositories. I feel like we're moving to a solution where the daemon is "blockchain agnositic", in a sense that it bases it's utxo set and balances only on cache, without relying on data fetched from the chain.

I think that instead of continuing to split storages, it could be worth evaluating using one for all of them (except the Unspent one that will in-memory).

altafan:
ACK

sekulicd:
@tiero @altafan check [comment](https://github.com/TDex-network/tdex-daemon/issues/305#issuecomment-801914183) related to complete solution.

PR #311, opened 2021-03-17
Hotfixes

altafan:
With the latest PRs we introduced some bugs and also left new methods untested. This might seem too heavy, but it's intended to consolidate domains in order to have them properly tested as the main goal.

For this reason, the recently added methods which were untested, have been removed, preferring instead to focusing on consolidating what we already had in place.

Methods removed:
- `MarketRepository.GetNonFundedMarkets()` in favour of delegating to app services to filter all markets and retrieve those non funded (only the operator requires this).
- `Vault.AddressesBlindingKeysGroupByAccount` in favour of uniforming all domain's methods starting with `All` to return a list of AddressInfo fro which the list of addresses or addresses and keys can be easily retrieved.

Hotfixes:
- `TradeService.TradePropose`  was failing to create a new Trade: it's been fixed by calling `TradeRepository.GetOrCreateTrade()` before `UpdateTrade` (it used to behave like an _upsert_ but recently became a pure _update_)
- `TradeService.TradeComplete` was always trying to set the trade failed. the args used to be concrete stub types which were easy to check if nil. Recently those args became interfaces that are harder to check if `nil`. It's been fixed by using pointer to interfaces to easily check if nil at app level. This makes the interface layer a bit harder to read but the decision is intended to prioritize a cleaner business logo against the presentation one.

Please @tiero review this.


tiero:
As minor note, Would have been better to make those comments on previous merged PR, instead of changing them.

ACK for hotfixes

Hope @sekulicd dosen't get offended :P ;)

tiero (approved):

sekulicd (approved):

PR #312, opened 2021-03-18
Remove addresses observation

### LINKED ISSUE #279: [Badger] Move unspents to in-memory mode

### LINKED ISSUE #293: [Proposal] Re-think the deposit design to not depend on blockchain polling
This is a competing idea with other that go in the line to increase the reliability of the daemon, which is not stable yet 

## Issues:

- Balances are not always in sync https://github.com/TDex-network/tdex-daemon/issues/289
- Maybe this also due crawling mechanics https://github.com/TDex-network/tdex-daemon/issues/290
- High CPU usage (partly due the handleUnspents) https://github.com/TDex-network/tdex-daemon/issues/287

## Competing proposals

- [Blockchain data fetching] Proposal: Expect external service to notify daemon about wallet activites  https://github.com/TDex-network/tdex-daemon/issues/260

Not entirely a drop-in alternative, but if #260 gets traction and is the way to go, the curren crawler + blockchain listener will be detached by the tdex-daemon, but eventually suffer from the same problem I guess (unless we discover that balances problems are due internal repositories, rather than actual blockchain polling)

- Smarter request throttling  https://github.com/TDex-network/tdex-daemon/issues/276

This would not need any more throttling

### TL;DR

Expect end user to "claim" deposits instead of observing addresses returned via RPC and update internal unspents view only at startup and after a "write" (ie. a Trade or a Witdrawal)

### Long (?) story

We need to keep a view of the unspents of the wallet in order to: 

- accept trades automatically with minimal latency
- serve Balances endpoint for the traders to know about reserves
- Eventually log operator about low balance for a market or the fee account

At the moment we have in place polling activity ie. every 5 seconds to know about if wallet state is changed. Altough we could increase to 1 min (Liquid block time), we would loose reponsiveness for future trades, halt concurrent trading for too long and as per curren deposit mechanics we would wait too much to fund a market.

To lower the polling activity we could assume the wallet state changes:

a) after a Trade succesful
b) after a Witdrawal 
c) after a new deposit
d) external wallet softawre using the same mnemonic (Is that necessary?)

- Excluding `d` wich def need polling, we could accomplish a-c without involing any blockchain polling:
- For `a` and `b` we update the internal cache right away with internal data after tx got broadcasted. We could make extra call (non blocking in async) to go check if something wrong, but this is just we are paranoid.
- For `c` the idea is to change the current RPC flow for the operator in a way is his duty to call an additional RPC ie. `ClaimDeposit` passing us txid(s) in a way we could go check to the chain when requested if the deposit is there and got confirmations. We could even asks for vout, this will also indicate from the operator wich market is wich (ie no inference)

### Motivation

Looking at the actual operations of a Liquidity Provider, the process to deposit is quite involving already and needs some dedication: a) source funds b) fragment them c) ask for and address and make push the transaction
Asking the LP to do an extra step to proivde us proof of the deposit, I see it not that more demanding. Also is something you do with max focus most of the time, you are going to have few moments of deposit, in not something frequent or casual activity.

### FAQ

TBD

 
﻿

### LINKED ISSUE #297: Do not observe an address when returned via DepositMarket and DepositFeeAccount
this should be done after #296, since the way to update internal Market is dependant on operator to claim specific transactions in an explicit way﻿

### LINKED ISSUE #298: Update "spentness" after RPC calls to ClaimMarketDeposit, ClaimFeeDeposit, WithdrawMarket & TradeComplete

### LINKED ISSUE #303: TradeComplete: respond ASAP to clients and update Trades repo in async

### LINKED ISSUE #310: Between TTL of locked unspents and eventual spentness update there could be window to coinselect an already spent tx

altafan:
This makes the daemon able to update its unspents while processing trades, (almost) without relying on the work of the blockchain listener.

### Operator service

The` ClaimMarketDeposit` and `ClaimFeeDeposit` RPCs have been updated so that the daemon, after ensuring that the provided deposits are correct, takes care of adding the unspents to the storage through the mean of non blocking go routines.
Within the `WithdrawMarketFunds` RPC, instead, after the tx gets broadcasted, the daemon extracts the unspents from it, and marks some of them SPENT, while some other is be added CONFIRMED.. respectively those used as inputs and the eventual change outputs generated.

### Wallet service

`SendToMany` RPC behavior is similar to the one of `WIthdrawMarketFunds` but focused only on updating unspents for the fee account.

### Trade service

The `TradePropose` RPC have been updated so that after accepting a trade proposal it makes the bc listener watching for  ONLY the inclusion of the trade tx gets in the chain.

After `TradeComplete` successfully completes a trade, it updates the utxo set by setting those used as inputs SPENT and by adding those created as output as UNCONFIRMED. It is the bc listener in charge of making them confirmed.

### Blockchain listener

This service now gets notified by the crawler ONLY when a trade transactions is included in the blockchain. When this happens, it settles the trade and confirms the unspents added in case the trade was already completed,  otherwise it takes care of adding the new ones (already confirmed this time), and mark those used as spent.

### Badger repo

With this, we don't use badger's native TTLs to unlock the unspents anymore. This is something that is done at application level by the trader service through a go routine now.
Unspent repo has been changed to InMemory. 

Closes #279.
Closes #293.
Closes #297.
Closes #298.
Closes #303.
Closes #310.

Please @tiero review this.

EDIT:
This makes the HTTP request timeout configurable while previously was hardcoded to 3 secs and was source of problems when using Esplora explorer.

Also, this introduces the usage of circuit breaker pattern to retry failing HTTP requests when fetching/restoring the utxo set.

These changes are required in order to fix the restore option of the InitWallet RPC.

BONUS:  This fixes ListUtxos RPC that was returning incorrect responses by including actually spent unspents in the _Locks_ list instead of the _Spents_ one.


tiero:
@sekulicd it seems #279 has been taken by @altafan  already

internal/core/application/trade_service.go
```diff
@@ -540,16 +570,72 @@ func (t *tradeService) getUnspentsBlindingsAndDerivationPathsForAccount(
...
+func (t *tradeService) checkTradeExpiration(
+	tradeTxID string,
+	selectedUnspentKeys []domain.UnspentKey,
+) {
+	// admit a tollerance of 1 minute past the expiration time.
+	time.Sleep(time.Duration(t.expiryDuration+60) * time.Second)
+	ctx := context.Background()
+
+	// if the trade is expired it's required to unlock the unspents used as input
+	// and to bring the trade to failed status
```
tiero, line 609: not sure about that: a trade should be expired status, it's not failed per se IMHO

internal/infrastructure/storage/db/badger/db_manager.go
```diff
@@ -201,8 +201,12 @@ func EncodeKey(key interface{}, typeName string) ([]byte, error) {
 func createDb(dbDir string, logger badger.Logger) (*badgerhold.Store, error) {
```
tiero, line 201: add comment that explain this implicit behavior

internal/core/application/blockchain_listener.go
```diff
@@ -93,6 +99,7 @@ func newBlockchainListener(
 		pendingObservables:  make([]crawler.Observable, 0),
 		marketBaseAsset:     marketBaseAsset,
 		feeBalanceThreshold: feeBalanceThreshold,
+		network:             net,
```
tiero, line 102: Except for trade and unspent epo, are other repos necessary if we only watch txs now?

tiero (changes requested):
We may need to populate the unspent repo after each unlock/startup

tiero (changes requested):
After an initial --restore && unlock, if you stop and restart the daemon it will get stuck with never reaching the point that can starts listening on http interfaces

ie.

```
time="2021-03-18T23:14:39Z" level=info msg="Replaying file id: 0 at offset: 0\n"
time="2021-03-18T23:14:39Z" level=info msg="Replay took: 274.474µs\n"
time="2021-03-18T23:14:39Z" level=info msg="All 0 tables opened in 0s\n"
time="2021-03-18T23:14:39Z" level=info msg="Replaying file id: 0 at offset: 0\n"
time="2021-03-18T23:14:39Z" level=info msg="Replay took: 6.118µs\n"
time="2021-03-18T23:14:52Z" level=info msg="All 0 tables opened in 0s\n"
time="2021-03-18T23:14:52Z" level=info msg="Replaying file id: 0 at offset: 0\n"
time="2021-03-18T23:14:52Z" level=info msg="Replay took: 275.879µs\n"
time="2021-03-18T23:14:52Z" level=info msg="All 0 tables opened in 0s\n"
time="2021-03-18T23:14:52Z" level=info msg="Replaying file id: 0 at offset: 0\n"
time="2021-03-18T23:14:52Z" level=info msg="Replay took: 6.586µs\n
```

```
[tdex] rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp [::1]:9000: connect: connection refused"

```

tiero:
It seems latest commit in master introduced some bug/regression

I loaded the same mnemonic locally 
```
tdex-daemon refactor-trade-complete ✅  DEX_LOG_LEVEL=5 TDEX_CRAWL_LIMIT=250 TDEX_CRAWL_TOKEN=10 TDEX_CRAWL_INTERVAL=5000 ./build/tdexd-linux-amd64
INFO[0000] All 0 tables opened in 0s                    
INFO[0000] Replaying file id: 0 at offset: 0            
INFO[0000] Replay took: 81.332µs                        
INFO[0000] All 0 tables opened in 0s                    
INFO[0000] Replaying file id: 0 at offset: 0            
INFO[0000] Replay took: 1.283µs                         
INFO[0000] All 0 tables opened in 0s                    
INFO[0000] Replaying file id: 0 at offset: 0            
INFO[0000] Replay took: 216.173µs                       
INFO[0031] starting daemon                              
INFO[0031] trader interface is listening on :9945       
INFO[0031] operator interface is listening on :9000     
INFO[0068] fee account funded. Trades can be served     
INFO[0081] funding market with quote asset ce091c998b83c78bb71a632313ba3760f1763d9cfcffae02258ffa9865a37bd2 
WARN[0081] market not found                             
INFO[0082] funding market with quote asset ce091c998b83c78bb71a632313ba3760f1763d9cfcffae02258ffa9865a37bd2 
WARN[0082] market not found                             
INFO[0082] funding market with quote asset ce091c998b83c78bb71a632313ba3760f1763d9cfcffae02258ffa9865a37bd2 
WARN[0082] market not found                             
INFO[0082] funding market with quote asset ce091c998b83c78bb71a632313ba3760f1763d9cfcffae02258ffa9865a37bd2 
WARN[0082] market not found                             
INFO[0082] funding market with quote asset ce091c998b83c78bb71a632313ba3760f1763d9cfcffae02258ffa9865a37bd2 
WARN[0082] market not found  
```

altafan:
> After an initial --restore && unlock, if you stop and restart the daemon it will get stuck with never reaching the point that can starts listening on http interfaces

When restarting the daemon, the utxo set must be restored,  and it's done when initializing the wallet service, just like it was before this, with theedifference that now we have to make use of the explorer for this.
The more are the addresses derived by the daemon, the more is the waiting time to recontruct the utxo set.

This is why it seems to you the daemon gets stuck after a restart, but it's actually recreating the unspents store, therefore when you say:

> We may need to populate the unspent repo after each unlock/startup

This is already done. But the explorer needs to be improved. I mean, all the waiting time is due to one single call that is `explorer.GetUnspentsForAddresses`, where we pass literally all addresses of the wallet, and wait for the explorer to do fetch the unspents for any of them. This is done sequentially so far, but it could be enhanced through the use go routines and rate limiting I think.

internal/core/application/blockchain_listener.go
altafan: Yes. They're are all still used to retrieve info about a trade . MarketRepo is used to get the market account index given its quote asset, while VaultRepo is used to retrieve addresses info for some account (fee/market).

internal/core/application/trade_service.go
altafan: Ok I'm going to add a new `ExpiredStatus` at domain level, but we'll need also a new `Expire()` method in charge of updating the status accordingly. Just like we have `Propose()`, `Accept()`, and so on for other statuses.

internal/core/application/wallet_service.go
```diff
@@ -845,3 +868,167 @@ func getBalancesByAsset(unspents []explorer.Utxo) map[string]BalanceInfo {
 	}
 	return balances
 }
+
+func fetchUnspents(
```
sekulicd, line 872: fetchUnspent actually:
1. fetches utxos
2. converts utxos to unspnts
3. stors unspents

altafan: yes, just like you do in your unnamed function within RelaodUtxo. fetchUnspents expects a list of AddressInfo, so you would first retrieve all addresses info from the vault and then pass down the list to fetchUnspents.

sekulicd: i was thinking maybe to rename it since it is doing 3 things :)

sekulicd: just a suggestion

internal/core/application/trade_service.go
```diff
@@ -429,37 +424,69 @@ func (t *tradeService) tradeComplete(
...
+			log.Error("unable to persist completed trade with id ", trade.ID, " : ", err.Error())
+		}
 
-			log.Info("trade with id ", tradeID, " broadcasted: ", txID)
+		_, accountIndex, _ := t.marketRepository.GetMarketByAsset(
+			ctx,
+			trade.MarketQuoteAsset,
+		)
+
+		unspentsToAdd, unspentsToSpend, err := extractUnspentsFromTx(
```
sekulicd, line 472: @altafan can we use here extractUnspentsFromTxAndUpdateUtxoSet method?

altafan: done

tiero:
Can you QA test with me this PR also? @sekulicd Especially now that we have the ListUtxos and ReloadUtxos endpoint to "debug" "it

sekulicd:
> Can you QA test with me this PR also? @sekulicd Especially now that we have the ListUtxos and ReloadUtxos endpoint to "debug" "it

Are latest changes from master merged to this branch?

tiero:
not yet I do believe. Woudl you push to altafan branch please? @sekulicd

sekulicd:
@altafan i am getting this error in logs:
ERRO[9000] Cannot run value log GC when DB is opened in InMemory mode

altafan:
>@altafan i am getting this error in logs:
ERRO[9000] Cannot run value log GC when DB is opened in InMemory mode

Fixed.

tiero:
can you solve conflicts @altafan

Issue #254, opened 2021-02-18
Refactor Testing


tiero:
This is needed for tracking.

Read the https://github.com/TDex-network/tdex-daemon/pull/219#issuecomment-732092280 thread for reference and context.

## TODOs

- [x] Refactor domains
- [x] add unit tests for domains
- [x] add component tests
- [ ] rewrite the e2e test

altafan:
@tiero link doesn't work.

tiero:
> @tiero link doesn't work.

Updated

PR #315, opened 2021-03-19
Rpc reload utxos

### LINKED ISSUE #299: Add ReloadUtxos RPC to allow operator to trigger the daemon to update programmatically the utxo set
Once we have #297 #296  #298, it could help the operator to be able to recover the eventuality the balances are not up to date for some internal error.

This would make use of the explorer to go check the unspents of all derived addresses  and to update the local storage.

```
service Operator {
    rpc ReloadUtxos(ReloadUtxosRequest) returns(ReloadUtxosReply)
}

message ReloadUtxosRequest {}
message ReloadUtxosReply {}
```

sekulicd:
This adds ReloadtUtxos RPC implementation which can be used to reload unspents from blockchain.

This closes #299 

### DIFF
--- a/internal/core/application/operator_service.go
+++ b/internal/core/application/operator_service.go
@@ -88,6 +88,7 @@ type OperatorService interface {
 		ctx context.Context,
 		market Market,
 	) (*ReportMarketFee, error)
+	ReloadUtxos(ctx context.Context) error
 }
 
 type operatorService struct {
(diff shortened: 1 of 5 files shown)


internal/interfaces/grpc/handler/operator.go
```diff
@@ -146,6 +146,28 @@ func (o operatorHandler) ReportMarketFee(
 	return o.reportMarketFee(ctx, req)
 }
 
+func (o operatorHandler) ReloadUtxos(
+	ctx context.Context,
+	rew *pb.ReloadUtxosRequest,
+) (*pb.ReloadUtxosReply, error) {
+	res, err := o.dbManager.RunTransaction(
```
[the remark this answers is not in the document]
tiero: so it's not a real conflict, rather than "db is busy".  In this scenario seems to me that is acceptable to have a **stale read** if the operator is reloading utxos set in the same moment is deriving a new address or a trade is deriving a new change address, is ok to me to reload the "old" list of address only.

but due snapshot isolation of badger it seems to me that this is not possible right? to "force" a stale read I mean.

internal/core/application/operator_service.go
```diff
@@ -836,6 +837,61 @@ func (o *operatorService) FeeAccountBalance(ctx context.Context) (
...
+		nil,
+		"",
+		nil,
+	)
+	if err != nil {
+		return err
+	}
+
+	addressesInfo := vault.AllDerivedAddressesInfo()
+	for _, v := range addressesInfo {
```
altafan, line 854: in #312 a new function fetchUnspents() has been added that already that this work. If we wait for that PR to be merged, this portion of code could be simplified as a simple function call.

internal/interfaces/grpc/handler/operator.go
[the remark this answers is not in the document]
sekulicd: > but due snapshot isolation of badger it seems to me that this is not possible right? to "force" a stale read I mean.

i dont think it is possible to read stale data

> The conflict that can arise here just if it fails to acquire the lock on vault repo to retieve the Vault right?

I think that it transaction conflict will be thrown if vault read happen before some other tx updates vault.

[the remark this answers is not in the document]
tiero: gotcha 👍  although this seems a read-only tx, should not interfer that much https://dgraph.io/docs/badger/get-started/#read-only-transactions

internal/core/application/operator_service.go
tiero: fetchUnspents lives in the wallet_service isn't it? should we export it or expose isnt'

altafan: it is a private function of the `application` package, therefore the operator service can use it since it leaves in the same package.

tiero: you right nvm

tiero: I woudl either ask @sekulicd to cherrypick or copy paste that method, since your PR is bigger and will go inside provably next week after a bit of more QA

tiero (approved):
utACK 

altough no tests (we will remand later with #254) lgtm

PR #317, opened 2021-03-19
Fix wallet restore

altafan:
This starts from #312 and fixes the restoration of the wallet.

@tiero do you want to test this out?

I made several tries to check if all works as expected. For each of the following situations I tried to complete the flow in order to broadcast at least 2 trades (1 BUY and 1 SELL):
* new wallet from scratch
* restart (start-stop-start) with already used wallet 
* restore already used wallet
* restart from restored wallet (start-with-restore-stop-start)

This makes the HTTP request timeout configurable while previously was hardcoded to 3 secs and was cause of problems when using Esplora explorer.

Also, this introduces the usage of circuit breaker pattern to retry failing HTTP requests when fetching/restoringthe utxo set.

~~What I noticed:
sometimes (but it might actually  be everytime), with a new wallet from scratch, the TradePropose RPC fails with some `Invalid prefix...` error. When that happens I just restart the daemon so it re-builds the utxo set and everything works fine then. Would be nice if someone replicated this. I suspect the problem could come from the ClaimDeposit RPC(s)~~

EDIT: fixed.


tiero:
as discussed offline we must:

a) make the timeout ot the httputil package configurable and increase to something like 15 seconds
b) add a circuit breaker meachism with a reasonable number of max requests, before declare Esplora in faulty state

tiero:
Would you update the PR description?

internal/core/application/wallet_service.go
```diff
@@ -880,13 +1123,176 @@ func fetchUnspents(
...
+}
+
+func spendUnspents(
+	unspentRepo domain.UnspentRepository,
+	unspentKeys []domain.UnspentKey,
+) (int, error) {
+	return unspentRepo.SpendUnspents(context.Background(), unspentKeys)
+}
+
+func addUnspentsAsync(unspentRepo domain.UnspentRepository, unspents []domain.Unspent) {
```
sekulicd, line 1157: why Async?

internal/core/application/wallet_service.go
```diff
@@ -880,13 +1123,176 @@ func fetchUnspents(
...
+		log.Warnf(
+			"unexpected error occured while adding unspents to the utxo set. "+
+				"You must manually run ReloadUtxo RPC as soon as possible to restore "+
+				"the utxo set of the internal wallet. Error: %v", err,
+		)
+	}
+	log.Debugf("added %d unspents", len(unspents))
+}
+
+func spendUnspentsAsync(
```
sekulicd, line 1168: why Async?

sekulicd:
> Invalid prefix... error. When that happens I just restart the daemon so it re-builds the utxo set and everything works fine then. Would be nice if someone replicated this. I suspect the problem could come from the ClaimDeposit RPC(s)

I also noticed this, definitely  looks like there is some issue with ClaimMarketDeposit or ClaimFeeDeposit, since when i run ReloadUtxo RPC it works fine.

sekulicd:
Also I think we are not doing transitions of uxtos properly, PSB bellow output of ListUtxos after one trade:

```
{
  "info_per_account": {
    "\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000": {
      "unspents": [
        {
          "outpoint": {
            "hash": "f96df2e95410dd3fb98db3f2a2cbc51a7d50979ea7f842ae787a220fbed69058",
            "index": 4
          },
          "value": "99998700",
          "asset": "5ac9f65c0efcc4775e0baec4ec03abdde22473cd3cf33c0419ca290e0751b225"
        }
      ],
      "spents": [],
      "locks": [
        {
          "outpoint": {
            "hash": "5611ffa3bc2ea1c71a30d311bd04c4419aa1572256b5450196ced3f3ae758089",
            "index": 1
          },
          "value": "100000000",
          "asset": "5ac9f65c0efcc4775e0baec4ec03abdde22473cd3cf33c0419ca290e0751b225"
        },
        {
          "outpoint": {
            "hash": "f858674f496331c4009fc2f08b8cef8794fe3ebbf84a586337d150c9ce9657cd",
            "index": 4
          },
          "value": "99999350",
          "asset": "5ac9f65c0efcc4775e0baec4ec03abdde22473cd3cf33c0419ca290e0751b225"
        }
      ]
    },
    "\u0005\u0000\u0000\u0000\u0000\u0000\u0000\u0000": {
      "unspents": [
        {
          "outpoint": {
            "hash": "360be2881000e6473cd2a9f71204e60e85778f65dbda558cf5ddd9ea4a54a260",
            "index": 0
          },
          "value": "100000000",
          "asset": "5ac9f65c0efcc4775e0baec4ec03abdde22473cd3cf33c0419ca290e0751b225"
        },
        {
          "outpoint": {
            "hash": "f858674f496331c4009fc2f08b8cef8794fe3ebbf84a586337d150c9ce9657cd",
            "index": 3
          },
          "value": "18458740671000",
          "asset": "f609439a7c41e2ebb580f8ebbf14b92e56939de4c71dc6abd9f0dd6f3dea45ab"
        },
        {
          "outpoint": {
            "hash": "f96df2e95410dd3fb98db3f2a2cbc51a7d50979ea7f842ae787a220fbed69058",
            "index": 2
          },
          "value": "876795701805",
          "asset": "f609439a7c41e2ebb580f8ebbf14b92e56939de4c71dc6abd9f0dd6f3dea45ab"
        },
        {
          "outpoint": {
            "hash": "f96df2e95410dd3fb98db3f2a2cbc51a7d50979ea7f842ae787a220fbed69058",
            "index": 3
          },
          "value": "5000000",
          "asset": "5ac9f65c0efcc4775e0baec4ec03abdde22473cd3cf33c0419ca290e0751b225"
        }
      ],
      "spents": [],
      "locks": [
        {
          "outpoint": {
            "hash": "3076c2cab70a33ac4127bc6396220882c2cea8a768cafc389eea1f469c3df16d",
            "index": 1
          },
          "value": "20300000000000",
          "asset": "f609439a7c41e2ebb580f8ebbf14b92e56939de4c71dc6abd9f0dd6f3dea45ab"
        },
        {
          "outpoint": {
            "hash": "f858674f496331c4009fc2f08b8cef8794fe3ebbf84a586337d150c9ce9657cd",
            "index": 2
          },
          "value": "10000000",
          "asset": "5ac9f65c0efcc4775e0baec4ec03abdde22473cd3cf33c0419ca290e0751b225"
        }
      ]
    }
  }
}
```

internal/core/application/operator_service.go
```diff
@@ -961,57 +988,105 @@ func (o *operatorService) claimDeposit(
...
-		for _, s := range outpointScripts {
-			if _, ok := accountScripts[s]; ok {
-				count++
+			unspents[i] = domain.Unspent{
+				TxID:            tx.TxHash().String(),
+				VOut:            uint32(v.Index),
+				Value:           unconfidential.Value,
+				AssetHash:       unconfidential.AssetHash,
+				ValueCommitment: bufferutil.CommitmentFromBytes(txOut.Value),
+				AssetCommitment: bufferutil.CommitmentFromBytes(txOut.Value),
```
sekulicd, line 1049: should be txOut.Asset?

sekulicd: yeap, this is the issue with 'Invalid prefix issue' happening cause wrong value for AssetCommitment.

internal/core/application/wallet_service.go
altafan: because it handle errors and logs/warns by itself since it's supposed to be called by a go routine

sekulicd: i understand your point, but if not called with go routine its actually sync

internal/core/application/operator_service.go
altafan: fixing it, thanks!

internal/core/application/wallet_service.go
altafan: yes I agree with you but I didn't find a better name for these methods.. do you have any suggestion?

altafan:
>as discussed offline we must:

>a) make the timeout ot the httputil package configurable and increase to something like 15 seconds
>b) add a circuit breaker meachism with a reasonable number of max requests, before declare Esplora in faulty state

Done. The default req timeout is now 15 secs and configurable via env var. The circuit breaker opens the circuit when more than 20 requests have been made with a failure ratio of the 70%. Is it fine?

config/config.go
```diff
@@ -26,6 +26,8 @@ const (
 	OperatorListeningPortKey = "OPERATOR_LISTENING_PORT"
 	// ExplorerEndpointKey is the endpoint where the Electrs (for Liquid) REST API is listening
 	ExplorerEndpointKey = "EXPLORER_ENDPOINT"
+	// ExplorerRequestTimeoutKey is the max time to wait for HTTP responses before it expires
```
tiero, line 29:
```suggestion
	// ExplorerRequestTimeoutKey are the milliseconds to wait for HTTP responses before it timeouts
```

internal/core/application/errors.go
```diff
@@ -4,17 +4,19 @@ import "errors"
...
 	// ErrMarketNotExist ...
 	ErrMarketNotExist = errors.New("market does not exists")
+	// ErrMarketNotFunded ...
+	ErrMarketNotFunded = errors.New("market account not funded")
 	// ErrMissingNonFundedMarkets ...
 	ErrMissingNonFundedMarkets = errors.New("no non-funded markets found")
 	// ErrInvalidOutpoint ...
 	ErrInvalidOutpoint = errors.New("outpoint refers to inexistent tx output")
+	// ErrInvalidOutpoints ...
+	ErrInvalidOutpoints = errors.New("all outpoints must be funds of to the same account")
```
tiero, line 21:
```suggestion
	ErrInvalidOutpoints = errors.New("all outpoints must be funded for the same account")
```

internal/core/application/trade_service.go
```diff
@@ -540,16 +562,74 @@ func (t *tradeService) getUnspentsBlindingsAndDerivationPathsForAccount(
...
+
+	log.Debugf("unlocked %d unspents", count)
+}
+
+func (t *tradeService) checkTradeExpiration(
+	tradeTxID string,
+	selectedUnspentKeys []domain.UnspentKey,
+) {
+	// admit a tollerance of 1 minute past the expiration time.
+	time.Sleep(time.Duration(t.expiryDuration+60) * time.Second)
```
tiero, line 597: I do think is better to lift up this sleep so it's visible in the upper code

internal/core/application/validators.go
```diff
@@ -20,7 +20,7 @@ func validateAmount(satoshis decimal.Decimal) error {
 }
 
 func validateAssetString(asset string) error {
-	const regularExpression = `[0-9A-Fa-f]{64}`	
+	const regularExpression = `[0-9a-f]{64}`
```
tiero, line 23: why changing the expression?

internal/core/application/wallet_service_test.go
```diff
@@ -48,7 +49,16 @@ func TestInitWalletWrongSeed(t *testing.T) {
...
 
 	wrongSeed := []string{"test"}
-	err := walletSvc.InitWallet(ctx, wrongSeed, "pass", !restoreWallet)
+	chErr := make(chan error, 1)
+	walletSvc.InitWallet(
+		ctx,
+		wrongSeed,
+		"pass",
+		!restoreWallet,
+		make(chan *InitWalletReply),
```
tiero, line 58: would you put this in a variable above for better reading the parameter names?

internal/infrastructure/storage/db/badger/unspent_repository_impl.go
```diff
@@ -259,129 +257,165 @@ func (u unspentRepositoryImpl) getBalance(
...
+
+	if err := u.insertLock(ctx, key, tradeID); err != nil {
+		return false, err
 	}
 
-	return u.insertLock(ctx, key, tradeID)
+	return true, nil
 }
 
 func (u unspentRepositoryImpl) unlockUnspents(
```
tiero, line 380: It looks liek here we are dropping TTL therefore we could have used `badgerhold` abstractions. But maybe it's ok like this

internal/core/application/validators.go
altafan: asset hashes in hex string format should be all-lower-case we must not allow upper case ones.

internal/infrastructure/storage/db/badger/unspent_repository_impl.go
altafan: TTLs are not ok anymore because we need to need to update the lock and spent status together, otherwise there could be the chance that some actually spent utxos might be used for some trades just because the spent status is updated AFTER the trade proposal.

altafan: let's open a ticket for that.

tiero (changes requested):
Update `tdex initi` command

tiero (approved):

altafan:
Merged into #312. Closing.

PR #322, opened 2021-03-26
[WIP]Utxo fragmentation

sekulicd:
This is new PR which should replace #194.

It merges master on top of previous PR and adds claim unspents automatically.
It is in WIP since it needs to be better tested alongside with Wallet restore and other RPCs that could be affected with high number of generated addresses and unspents (eg. RestoreUtxos)

This closes #184 

@tiero @altafan please review.

PR #326, opened 2021-04-02
Component tests & Fixes

### LINKED ISSUE #325: Investigate transaction size estimation
From clients we obtain a lot of errors when broadcasting such as `min relay fee not met, 650 < 653 (code 66)\`

﻿

altafan:
This adds component test for wallet and operator services:
* for the former, tests validate the init wallet flow for a new wallet from scratch, a restarted daemon and a restored wallet.
* for the latter, there's a test that validates the flow of creating and funding fee/market accounts. As soon as we support blinding via blinders, we should add also a test to validate the withdrawal from market.

BONUS:
* Fix error messages of init wallet when it's not yet synced/initialized.
* DbManager has been renamed DbManager and repos are now created and retrieved through it.
* WalletService.DepositMarket now returns a slice of AddressAndBlindingKey instead of only addresses. It's responsibility of the grpc interface to return only addresses.
* Add `BlinderManager` to be able to mock unblinding of outputs
* Add GetTransaction to explorer.Service
* Add methods to blind with blinding data in pkg/wallet. TODO: use them to let the daemon blind txs without unblinding inputs again (it already has them as unblinded unspent).
* Drop MinRelayFee and refactor tx size estimation.

This closes #325.

Please @tiero review this.


altafan:
@sekulicd you may want to look at how `EstimateTxSize` has been changed.

pkg/wallet/estimation.go
```diff
@@ -110,13 +113,21 @@ func calcTxBaseSize(
...
 
-func calcTxWitnessSize(inScriptTypes, outScriptTypes []int) int {
+func calcTxWitnessSize(
+	inScriptTypes, inAuxiliaryWitnessSize,
+	outScriptTypes []int,
+) int {
 	insSize := 0
+	auxCount := 0
 	for _, scriptType := range inScriptTypes {
-		if scriptType > P2MS {
```
tiero: is this breaking any of other project may using this package?

altafan: the estimation function wasn't exposed before this or #322, so other projects would not be affected by this. Anyway, I changed this specific check to increase the precision of estimation for txs with ins/outs of type p2sh(p2wsh) or p2wsh.

tiero (approved):

pkg/wallet/util.go
```diff
@@ -161,9 +161,10 @@ func addInsAndOutsToPset(
 	return ptx.ToBase64()
 }
 
-func extractScriptTypesFromPset(ptx *pset.Pset) ([]int, []int, []int, []int) {
+func extractScriptTypesFromPset(ptx *pset.Pset) ([]int, []int, []int, []int, []int) {
```
sekulicd, line 164: should we move this to esitamtion.go ?

pkg/wallet/util.go
```diff
@@ -213,13 +222,26 @@ func extractScriptTypesFromPset(ptx *pset.Pset) ([]int, []int, []int, []int) {
...
 			outScriptTypes[i] = P2WSH
 		}
 	}
 
-	return inScriptTypes, outScriptTypes, inAuxiliaryP2ShSize, outAuxiliaryP2ShSize
+	return inScriptTypes, inAuxiliaryRedeemScriptSize, inAuxiliaryWitnessSize,
+		outScriptTypes, outAuxiliaryRedeemScriptSize
+}
+
+func calcWitnessSizeFromRedeemScript(script []byte) int {
```
sekulicd, line 237: should we move this to esitamtion.go ?

sekulicd (commented):
we should use camel case for constants also.

pkg/wallet/util.go
altafan: I put these functions in util because they're used in transasction.go

PR #330, opened 2021-04-14
Trade service component test & move badger transaction creation at app level

### LINKED ISSUE #305: Move db tx management to application level
Currently, the operation of trying to commit a db transaction is delegated to the grpc handlers.
So basically we put an high-level component like a grpc handler (that should be responsible just to translate data coming from a lower level into a specific format like protobuf, json, etc.) in charge of doing a low-level operation like re-trying to commit a db tx if it caused a conflict.

Everything related to storage management should live in the application layer. So, as the title suggests, these kind of operations should be moved at application level. This would make it possible to write component tests to validate the behavior of the app in case of concurrent requests.

### LINKED ISSUE #327: Remove proofs from Unspent model and use blinders

### LINKED ISSUE #328: Increase verbosity whenstarting the daemon about utxos reload

### LINKED ISSUE #329: feeBalanceThreshold not used anymore in bc listener

altafan:
This adds the component tests for the trade service, making sure that the daemon is able to serve multiple concurrent trade proposals.
I found out that for some reason the in-memory vault repo isn't trade-safe, therefore it wouldn't be possible to run trade tests in parallel. To quickly solve this, I opted for tweaking a bit the existing badger repo manager and allow to create all stores in memory. Pure in-memory repositories will be used again in tests when we'll move to a weak consistency approach eventually.

This also enforces strong consistency at the application level: basically, the repoManager has been removed from the grpc interface, where it was living only to run manual db transactions (strong consistency) when required, ie. whenever an application RPC causes cross repo changes. This operation is now done by the app services if required.

BONUS:
* remove range and surjection proof from unspent domain
* because of above, blind txs with blinding data instead of keys
* trade repo creates a trade if not found (both impls)
* return blinding data also in case of unconf output in `trasactionutil.UnblindOutput`
* remove checks in pkg/swap involving unblinding of tx ins/outs.
* increase verbosity of wallet service when is restoring utxo set after daemon restart
* polish bc listener removing all its deprecated fields

Closes #305.
Closes #327.
Closes #328.
Closes #329.

Please @tiero review this.

### DIFF
--- a/internal/core/application/operator_service.go
+++ b/internal/core/application/operator_service.go
@@ -174,25 +174,47 @@ func (o *operatorService) DepositMarket(
 		numOfAddresses = 1
 	}
 
-	list := make([]AddressAndBlindingKey, numOfAddresses, numOfAddresses)
-	if err := o.repoManager.VaultRepository().UpdateVault(
-		ctx,
-		func(v *domain.Vault) (*domain.Vault, error) {
-			for i := 0; i < numOfAddresses; i++ {
-				info, err := v.DeriveNextExternalAddressForAccount(accountIndex)
-				if err != nil {
-					return nil, err
-				}
-				list[i] = AddressAndBlindingKey{
-					Address:     info.Address,
-					BlindingKey: hex.EncodeToString(info.BlindingKey),
-				}
-			}
-
-			return v, nil
-		}); err != nil {
+	vault, err := o.repoManager.VaultRepository().GetOrCreateVault(ctx, nil, "", nil)
+	if err != nil {
 		return nil, err
 	}
+
+	list := make([]AddressAndBlindingKey, numOfAddresses, numOfAddresses)
+	for i := 0; i < numOfAddresses; i++ {
+		info, err := vault.DeriveNextExternalAddressForAccount(accountIndex)
+		if err != nil {
+			return nil, err
+		}
+		list[i] = AddressAndBlindingKey{
+			Address:     info.Address,
+			BlindingKey: hex.EncodeToString(info.BlindingKey),
+		}
+	}
+
+	go func() {
+		if _, err := o.repoManager.RunTransaction(ctx, false, func(ctx context.Context) (interface{}, error) {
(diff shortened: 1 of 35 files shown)


internal/core/application/operator_service.go
```diff
@@ -171,43 +168,53 @@ func (o *operatorService) DepositMarket(
 		}
 
 		nextAccountIndex := latestAccountIndex + 1
-		fee := o.marketFee
```
tiero, line 174: can you explain this deletion?

internal/core/application/operator_service.go
```diff
@@ -171,43 +168,53 @@ func (o *operatorService) DepositMarket(
...
+				&domain.Market{AccountIndex: accountIndex, Fee: o.marketFee},
+			); err != nil {
+				return nil, err
 			}
 
-			return v, nil
+			if err := o.repoManager.VaultRepository().UpdateVault(
+				ctx,
+				func(_ *domain.Vault) (*domain.Vault, error) {
+					return vault, nil
```
tiero, line 208: here we are not making any change to the Vault. Isn't?

internal/core/application/operator_service.go
altafan: o.marketFee is used only when invoking GetOrCreateMarket... so I found excessive declaring a variable for that and I removed it... now o.marketFee is directly used as arg of GetOrCreateMarket invocation

tiero: I meant the whole block of code

tiero (commented):
LGTM, want to understand functional change in deposit market

internal/core/application/operator_service.go
altafan: Since this RPC makes cross repo changes (to vault and possible market repos), all the write operations are required to be executed within a non-managed db tx to enforce strong consistency. For this reason, the RPC now sources the domains or create new ones, and after it's done its job, applies the changes to the repo in a tx managed by the repo manager..

PR #332, opened 2021-04-15
Utxo fragmentation

### LINKED ISSUE #184: Concurrent swaps: Fragment UTXOS

altafan:
This finishes the work started with #194 and continued with #322.

This fixes conflicts and errors due to syncing with master. Taking in consideration what done in #322 this adds the following features:
* blind txs with blinding data
* force the operator to pass the txid(s) of the funds sent to the ephemeral address
* hotfix to trades expiry duration in main.go

EDIT: this adds new commands `fragmentfee` and `fragmentmarket` that do the fragmentation flow instead of adding it to the `depositfee` and `depositmarket`. If for any reason the flow fails, it's possible to resume that by providing the txids of the funds already sent to the ephemeral wallet, which is now persisted until the flow is completed. After that, it is flushed and a new one will be created for another `fragment` operation.

Closes #184.

Please @tiero, @sekulicd review this.


cmd/tdexd/main.go
```diff
@@ -62,12 +62,13 @@ func main() {
...
 	marketsFee := int64(config.GetFloat(config.DefaultFeeKey) * 100)
 	marketsBaseAsset := config.GetString(config.BaseAssetKey)
-	tradesExpiryDuration := config.GetDuration(config.TradeExpiryTimeKey)
+	tradesExpiryDuration := config.GetDuration(config.TradeExpiryTimeKey) * time.Second
 	withElementsSvc := config.IsSet(config.ElementsRPCEndpointKey)
 	pricesSlippagePercentage := decimal.NewFromFloat(config.GetFloat(config.PriceSlippageKey))
 	network := config.GetNetwork()
 	feeThreshold := uint64(config.GetInt(config.FeeAccountBalanceThresholdKey))
 
+	fmt.Println("tradesExpiryDuration", tradesExpiryDuration)
```
tiero, line 71: remove

cmd/tdexd/main.go
```diff
@@ -62,12 +62,13 @@ func main() {
 
 	marketsFee := int64(config.GetFloat(config.DefaultFeeKey) * 100)
 	marketsBaseAsset := config.GetString(config.BaseAssetKey)
-	tradesExpiryDuration := config.GetDuration(config.TradeExpiryTimeKey)
+	tradesExpiryDuration := config.GetDuration(config.TradeExpiryTimeKey) * time.Second
```
tiero, line 65: afaik GetDuration returns already a time.Duration

better renames the variable in what it is so `tradesExpiryDurationInSeconds`

cmd/tdex/depositmarket.go
```diff
@@ -2,26 +2,53 @@ package main
...
+	MinBaseDeposit     = 50000
+	MinQuoteDeposit    = 50000
+)
+
 var depositmarket = cli.Command{
 	Name:  "depositmarket",
 	Usage: "get a deposit address for a given market or create a new one",
 	Flags: []cli.Flag{
 		&cli.StringFlag{
-			Name:  "base_asset",
```
tiero, line 17: why we remove the base_asset

cmd/tdex/depositmarket.go
```diff
@@ -31,29 +58,534 @@ var depositmarket = cli.Command{
...
-		context.Background(),
-		&pboperator.DepositMarketRequest{
-			Market: &pbtypes.Market{
-				BaseAsset:  ctx.String("base_asset"),
-				QuoteAsset: ctx.String("quote_asset"),
+	net, err := getNetworkFromState()
+	if err != nil {
+		return err
+	}
+	baseAssetKey := net.AssetID
```
tiero, line 86: we are assuming base asset is always bitcoin, wich is not true.

Better be more explicit here and let the user full specify base & quote asset

cmd/tdex/depositmarket.go
```diff
@@ -31,29 +58,534 @@ var depositmarket = cli.Command{
...
-	printRespJSON(resp)
+	log.Info("calculating fragments...")
+	baseFragments, quoteFragments := fragmentUnspents(
+		assetValuePair,
+		fragmentationMapConfig,
+	)
+	numUnspents := len(unspents)
+	numFragments := len(baseFragments) + len(quoteFragments)
+	log.Infof(
+		"fetched %d funds that will be split into %d fragments",
```
tiero, line 144:
```suggestion
		"detected %d coins that will be split into %d fragments",
```

cmd/tdex/depositmarket.go
```diff
@@ -31,29 +58,534 @@ var depositmarket = cli.Command{
...
+		},
+	); err != nil {
+		return err
+	}
+
+	log.Info("done")
 	return nil
 }
+
+func fetchMarketAddresses(
```
tiero, line 216:
```suggestion
func getMarketDepositAddresses(
```

cmd/tdex/depositfee.go
```diff
@@ -27,17 +55,306 @@ func depositFeeAction(ctx *cli.Context) error {
...
+				baseAssetKey,
+				baseTotalAmount,
+			)
+		}
+	}
+
+	return valuePerAsset[baseAssetKey], unspents, nil
+}
+
+func fetchFeeAccountAddresses(
```
tiero, line 256:
```suggestion
func getFeeAccountAddresses(
```

cmd/tdex/depositfee.go
```diff
@@ -1,23 +1,51 @@
...
 	Flags: []cli.Flag{
+		&cli.BoolFlag{
+			Name:  "no_fragments",
+			Usage: "disable utxo fragmentation",
+			Value: false,
+		},
+		&cli.StringFlag{
+			Name:  "explorer",
+			Usage: "explorer endpoint url",
+			Value: "http://127.0.0.1:3001",
```
tiero, line 42: put the liquid blockstream one as default, for regtest we will have to pass

cmd/tdex/depositfee.go
```diff
@@ -1,23 +1,51 @@
...
+		},
 		&cli.IntFlag{
```
tiero, line 44: with fragmentation, this is of no use, am I right?

cmd/tdex/depositfee.go
```diff
@@ -27,17 +55,306 @@ func depositFeeAction(ctx *cli.Context) error {
...
+				out.SurjectionProof,
+				tx.Confirmed(),
+			))
+		}
+	}
+
+	return utxos, nil
+}
+
+func waitForOperatorFunds() []string {
```
tiero, line 355: it happens that if the txid is not found (slow to appear on mempool) and the user enter right away the txid it will say `tdex] Transaction not found` and exit the process

tiero: if the user call again depositfee it will prompt a different address

tiero: bets way would be to start a ticker to go check that txid until it founds, maybe with a reasonable timeout (1 minute or so)

cmd/tdex/fragmentfee.go
```diff
@@ -0,0 +1,373 @@
...
+			))
+		}
+	}
+
+	return utxos, nil
+}
+
+func waitForOperatorFunds() []string {
+	reader := bufio.NewReader(os.Stdin)
+	log.Info("Enter txid of fund(s) separated by a blank space: ")
```
tiero, line 366:
```suggestion
	log.Info("Enter txid of fund(s) separated by a white space: ")
```

cmd/tdex/depositfee.go
altafan: @tiero This flag has been removed. The explorer is always sourced from state and if it's not set via `tdex-cli config set explorer_url http://localhost:3001`, for example, it defaults to the liquid blockstream one.

cmd/tdex/fragmentfee.go
```diff
@@ -0,0 +1,373 @@
...
+		return err
+	}
+
+	if walletKeys == nil && txids != nil {
+		log.Info("no ephemeral wallet detected, skipping resume of provided txids")
+	}
+	if walletKeys != nil && txids == nil {
+		return fmt.Errorf(
+			"expected to resume previous fragmentation but no txids were provided." +
+				" Please retry by specifying txids with --txids",
```
tiero, line 75:
```suggestion
				" Please retry by specifying txids with --txid",
```

cmd/tdex/fragmentfee.go
```diff
@@ -0,0 +1,373 @@
...
+
+	walletKeys, err := getWalletFromState(walletType)
+	if err != nil {
+		return err
+	}
+
+	if walletKeys == nil && txids != nil {
+		log.Info("no ephemeral wallet detected, skipping resume of provided txids")
+	}
+	if walletKeys != nil && txids == nil {
```
tiero, line 72: I would log again here the address we have in storage

tiero (approved):

PR #334, opened 2021-04-20
Let the operator to configure CLI datadir

altafan:
This lets the user of the CLI to define its datadir by setting the env variable `TDEX_OPERATOR_DATADIR`. If not set it defaults to the usual appdata/tdex-operator.

Please @tiero review this.

### DIFF
--- a/cmd/tdex/main.go
+++ b/cmd/tdex/main.go
@@ -31,7 +31,7 @@ var (
 )
 
 func init() {
-	dataDir := os.Getenv("TDEX_CLI_DATADIR")
+	dataDir := os.Getenv("TDEX_OPERATOR_DATADIR")
 	if len(dataDir) <= 0 {
 		return
 	}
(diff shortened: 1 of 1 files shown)


cmd/tdex/main.go
```diff
@@ -30,6 +30,19 @@ var (
 	statePath   = path.Join(tdexDataDir, "state.json")
 )
 
+func init() {
+	dataDir := os.Getenv("TDEX_OPERATOR_DATADIR")
+	if len(dataDir) <= 0 {
+		return
+	}
+
+	if !path.IsAbs(dataDir) {
```
tiero, line 39: mmm why? lets resolve if is relative

altafan: I replicated the same behavior of btcd and lnd CLI's.. the path is expanded if starting with `~`. The relative path is allowed but not expanded.

tiero (approved):

PR #335, opened 2021-04-21
Hotfix: Restore proofs

altafan:
The last commit introduced a bug in the trade flow, and traders using the mobile app were experiencing an issue when the completing a swap. With that commit, the daemon didn't store range and surjection proofs of unspents anymore, causing the upset inputs of the swaps to have null proofs. This is likely to be the cause of the bug. The app, which uses liquidjs, is able to deserialize empty proofs, but it doesn't properly encode them in case they're empty.

This addresses the problem by restoring the proofs in the unspent domain, but keeping them always zeroed ([0x00]), as it's done in the Elements protocol for an output nonce when the output itself is uncofnidential: the nonce is zeroed, that means it contains a single 0 byte, instead of being null.

I made sure that go and js clients are able to complete swaps on regtest network, please try to replicate the same by yourself.

Please @tiero review this.

### DIFF
--- a/internal/core/application/operator_service.go
+++ b/internal/core/application/operator_service.go
@@ -1002,6 +1002,8 @@ func (o *operatorService) claimDeposit(
 				AssetBlinder:    unconfidential.AssetBlinder,
 				ScriptPubKey:    txOut.Script,
 				Nonce:           txOut.Nonce,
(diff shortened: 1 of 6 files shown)


tiero (approved):

louisinger (approved):
tACK

PR #347, opened 2021-04-26
Increase verbosity of FillProposal

altafan:
Please @tiero review this.


tiero (changes requested):
Can we log in the daemon console? Debug as log level it's ok

PR #350, opened 2021-04-28
Move to new ListTrade RPC

### LINKED ISSUE #292: [ListSwaps] Improve the Response
We need to change the ListSwaps method to be more useful:

* Add Tx Hash along with an Esplora like query string with blinders included
* Add Trade ID as well to match with logs
* Add SwapFail message if present

altafan:
This replaces the old operator service's ListSwaps RPC with the new ListTrades one.

In accordance with the [proto spec](https://github.com/TDex-network/tdex-protobuf/blob/master/operator.proto#L66) basically, new types have been added:
 * `SwapInfo` resuming info of the swap request message
 * `SwapFailInfo` resuming info of the swap fail message, if the trade failed
 * `TradeInfo` resuming info of a trade: besides those above, it contains:
   * the trade's id
   * the status including if the trade failed
   * the market (asset pair) it refers and the relative fee
   * both base and quote prices
   * in case the trade is settled, the Esplora URL of the tx including blinders to reveal the outputs
   * all timestamps of request, accept, complete, settle and exipry events in unix and UTC format

BONUS:
 * add the timestamp when completing a trade by calling trade.Complete(). This means that all the trades done 'till now by daemons outthere never recorded the completion timestamp.
 * fix wallet.ExtractBlindingDataFromTx in order to accept out blind keys and return also out blinding data besides those of the inputs.
 * add a ExtractBlindingData func to the application.TransactionManager helper.

This closes #292.

Please @tiero review this.


tiero (commented):
Could we sort swaps by date? maybe by request time?

altafan:
Done.

PR #351, opened 2021-04-29
General error in trade service & add balances to GetMarketPrice application method

### LINKED ISSUE #344: Vault domain error should not be lifted up to the Trade service

altafan:
This adds a generic error `service is unavailable, retry later` that is returned by the trade service in case of internal errors. The internal errors are logged rather than being forwarded to the user who made the request.

This closes #344.

Please @tiero review this.

### DIFF
--- a/internal/core/application/trade_service.go
+++ b/internal/core/application/trade_service.go
@@ -214,6 +214,16 @@ func (t *tradeService) GetMarketBalance(
 		return nil, domain.ErrMarketInvalidQuoteAsset
 	}
 
+	vault, err := t.repoManager.VaultRepository().GetOrCreateVault(ctx, nil, "", nil)
+	if err != nil {
+		log.Debugf("error while retrieving vault: %s", err)
+		return nil, ErrServiceUnavailable
+	}
+	if vault.IsLocked() {
+		log.Debug("vault is locked")
+		return nil, ErrServiceUnavailable
+	}
+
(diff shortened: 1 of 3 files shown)


internal/core/application/trade_service.go
```diff
@@ -168,30 +178,102 @@ func (t *tradeService) GetMarketPrice(
...
+	)
+	if err != nil {
+		log.Debugf("error while retrieving market: %s", err)
+		return nil, ErrServiceUnavailable
+	}
+	if accountIndex < 0 {
+		return nil, ErrMarketNotExist
+	}
+
+	vault, err := t.repoManager.VaultRepository().GetOrCreateVault(ctx, nil, "", nil)
```
tiero, line 229: better move this up as first check

internal/core/application/trade_service.go
```diff
@@ -211,19 +293,31 @@ func (t *tradeService) TradePropose(
...
 	if err != nil {
-		return nil, nil, 0, err
+		log.Debugf("error while retrieving market: %s", err)
+		return nil, nil, 0, ErrServiceUnavailable
 	}
 	if marketAccountIndex < 0 {
 		return nil, nil, 0, ErrMarketNotExist
 	}
 
+	vault, err := t.repoManager.VaultRepository().GetOrCreateVault(ctx, nil, "", nil)
```
tiero, line 303: same here

PR #357, opened 2021-05-04
Fix go routines

### LINKED ISSUE #352: Improve handling of go routines
We made pretty extensive use of goroutines so far to speed up the execution of the daemon's operation.

I took some time to review the correctness of these parts of code and it came up that there's room for improvements on how we handle concurrent code execution.

First of all, the main problem is that most of the concurrent code blocks make use of 2 channels, one for result(s) and one for error(s). I strongly suggest to use the [errgroup](https://pkg.go.dev/golang.org/x/sync/errgroup) package instead, which already
> provides synchronization, error propagation, and Context cancelation for groups of goroutines working on subtasks of a common task. 

This would increase performances and also code readability, hence maintainability.

The following is the list of functions that need to be fixed:
* [restoreUnspents](https://github.com/TDex-network/tdex-daemon/blob/master/internal/core/application/wallet_service.go#L757)
* [GetTransactionsForAddress (elements explorer)](https://github.com/TDex-network/tdex-daemon/blob/master/pkg/explorer/elements/transaction.go#L91)
* [toUtxos (elements explorer)](https://github.com/TDex-network/tdex-daemon/blob/master/pkg/explorer/elements/unspents.go#L94)
* [GetUnspentsForAddresses (esplora explorer)](https://github.com/TDex-network/tdex-daemon/blob/master/pkg/explorer/esplora/unspents.go#L19)
*  [getUnspents (esplora explorer)](https://github.com/TDex-network/tdex-daemon/blob/master/pkg/explorer/esplora/unspents.go#L43)

Another problem is in wallet service's InitWallet, where the channels are closed at [interface level](https://github.com/TDex-network/tdex-daemon/blob/04b9b990e695fd7d5d04b4ab44fa11efd4bcc5a1/internal/interfaces/grpc/handler/wallet.go#L93). They must be closed instead at application level, where data is wrote to channel, not there, where we read from the channel!

For this, it would be just enough to close the channels at app level without returning nil values, and the interface would know that channel has no more data to read by checking if it's been closed.

Last fix, the one with minor impact is to not use `select` at all when we read only from one channel. This happens in badger [repo manager](https://github.com/TDex-network/tdex-daemon/blob/04b9b990e695fd7d5d04b4ab44fa11efd4bcc5a1/internal/infrastructure/storage/db/badger/db_manager.go#L246) and [blockchain listener](https://github.com/TDex-network/tdex-daemon/blob/04b9b990e695fd7d5d04b4ab44fa11efd4bcc5a1/internal/core/application/blockchain_listener.go#L143).

altafan:
This addresses the handling of go routines in the codebase.

Instead of using errgroup as suggest in the related issue, I used waitgroups because the former allows only to retrieve errors while we want to manage return values also.

A 1 millisecond sleeptime has been added to those esplora methods that perform mulitple calls to prevent the latter to return error. Emprically, Esplora starts complaining (aka returning errors) if more than 340/400 concurrent requests arrive at the same time.

This also makes the channels in InitWallet RPC to be closed at application rather than interface level, simplifying the whole logic.

Use of `select` has been removed where not necessary.

Finally, some debug-level logs have been added to the wallet service to know how long restoration took to complete.

This closes #352.

Please @tiero @sekulicd  review this.


tiero (changes requested):

internal/core/application/wallet_service.go
```diff
@@ -752,18 +757,26 @@ func (w *walletService) getAllUnspentsForAccount(
...
+	wg := &sync.WaitGroup{}
+	wg.Add(len(info))
 
-	for _, in := range info {
+	go func() {
+		wg.Wait()
+		close(chUnspentsInfo)
+	}()
+
+	for i := range info {
```
sekulicd, line 778: I dont think we need waitgroups in here

```suggestion
	go func() {
	    defer close(chUnspentsInfo)
	    for i := range info {
		in := info[i]
		path := strings.Split(in.DerivationPath, "/")
		addressIndex, _ := strconv.Atoi(path[len(path)-1])
		chRes <- &InitWalletReply{
			AccountIndex: in.AccountIndex,
			AddressIndex: addressIndex,
			Status:       Processing,
			Data: fmt.Sprintf(
				"account %d index %d", in.AccountIndex, addressIndex,
			),
		}
		go w.restoreUnspentsForAddress(cb, in, chUnspentsInfo, wg)
		time.Sleep(1 * time.Millisecond)
	}
	}
```

internal/core/application/wallet_service.go
```diff
@@ -774,29 +787,28 @@ func (w *walletService) restoreUnspents(
 				"account %d index %d", in.AccountIndex, addressIndex,
 			),
 		}
-		go w.restoreUnspentsForAddress(cb, in, chUnspentsInfo, chErr)
+		go w.restoreUnspentsForAddress(cb, in, chUnspentsInfo, wg)
+		time.Sleep(1 * time.Millisecond)
+	}
```
sekulicd, line 792: go func() {
   defer close(chUnspentsInfo)
   for i := range info {
		in := info[i]
		path := strings.Split(in.DerivationPath, "/")
		addressIndex, _ := strconv.Atoi(path[len(path)-1])
		chRes <- &InitWalletReply{
			AccountIndex: in.AccountIndex,
			AddressIndex: addressIndex,
			Status:       Processing,
			Data: fmt.Sprintf(
				"account %d index %d", in.AccountIndex, addressIndex,
			),
		}
		go w.restoreUnspentsForAddress(cb, in, chUnspentsInfo, wg)
		time.Sleep(1 * time.Millisecond)
	}
        wg.Wait()
}

internal/core/application/wallet_service.go
```diff
@@ -752,18 +757,26 @@ func (w *walletService) getAllUnspentsForAccount(
...
+	wg.Add(len(info))
 
-	for _, in := range info {
+	go func() {
+		wg.Wait()
+		close(chUnspentsInfo)
+	}()
+
+	for i := range info {
+		in := info[i]
```
sekulicd, line 779:
```suggestion
	go func() {
		wg.Wait()
		close(chUnspentsInfo)
	}()

	for i := range info {
		in := info[i]
	go func() {
	    defer close(chUnspentsInfo)
	
```

sekulicd, line 779:
```suggestion
	go func() {
		wg.Wait()
		close(chUnspentsInfo)
	}()

	for i := range info {
		in := info[i]
	go func() {
	}
```

sekulicd, line 779:
```suggestion
	go func() {
		wg.Wait()
		close(chUnspentsInfo)
	}()

	for i := range info {
		in := info[i]
		
	go
```

pkg/explorer/elements/transaction.go
```diff
@@ -117,25 +123,27 @@ func (e *elements) GetTransactionsForAddress(addr string, blindingKey []byte) ([
...
-				}
-			case err := <-chErr:
-				close(chTxs)
-				close(chErr)
-				return nil, fmt.Errorf("tx details: %w", err)
+		chRes := make(chan txResult)
+		wg := &sync.WaitGroup{}
+		wg.Add(len(txids))
+
+		go func() {
```
sekulicd, line 130: My suggestion is, for the sake of readability is to have one anon func where we fill values in channel and close it with combination of wait group

````
go func() {
    defer close(chRes)
    for i := range txids {
	txid := txids[i]
	go e.getTxDetails(txid.(string), chRes, wg)
    }
    wg.Wait()
}
````

File: pkg/explorer/esplora/unspents.go [hunk 5]
```diff
@@ -59,120 +77,141 @@ func (e *esplora) getUnspents(addr string, blindingKeys [][]byte) ([]explorer.Ut
...
-					close(chUnspents)
-					return nil, fmt.Errorf("error on unblinding utxos: %s", err)
+		if len(blindingKeys) > 0 && r.utxo.IsConfidential() {
+			utxosToUnblind = append(utxosToUnblind, r.utxo)
+		} else {
+			utxos = append(utxos, r.utxo)
+		}
+	}
+
+	if len(utxosToUnblind) > 0 {
```
sekulicd, line 108: i think we are losing performance in here since we are waiting to get details for all utxo's and only then we do unblinding, we should be able to do unblinding as soon as there is 1 utxo with details

altafan: Consider that previously the execution was basically sequential even if we already use go routine s and channels... I don't think here we are loosing performance at all compared to what we have done before.

If you look at this form another perspective, If there's an error while getting the utxo details, we don't want to waste time in unblinding others because we return and abort all the process.

sekulicd: not comparing it to previous version, just saying that we are not as 'parallel' as we could.

pkg/explorer/elements/transaction.go
altafan: I might agree on the purpose, but making such suggested change causes the InitWallet w/ restore to fail and also the e2e test to fail as well sometimes.
Not really sure why, but appraently the way you're suggesting seems  to be less efficient, slowing down the execution.

Therefore, I won't commit the change.
If it's just for the sake of readability, since we handle this kind of operations always in the same way, we could reason about encapsulating the code in a sort of helper. I would open a separated ticket for it though.

pkg/explorer/esplora/unspents.go
altafan: I see.. I think I explained why we don't want to be as "concurrent" as we could though.

altafan:
@tiero in ci.yml I removed the testing-with-elements step since it's deprecated so far.
Originally, we used to switch between esplora and elements based on some env var for app tests, while now, in component tests, we make use of a mocked explorer.

PR #358, opened 2021-05-07
Move operator and wallet proto & Update services

### LINKED ISSUE #342: Move operator.proto api-spec defintion inside the daemon repository

### LINKED ISSUE #353: Fix type of application.Balance fields
The type of BaseAmount and QuoteAmount of the [application.Balance](https://github.com/TDex-network/tdex-daemon/blob/master/internal/core/application/types.go#L99) type must be `uint64` not `int64`.

### LINKED ISSUE #355: Add a FixedFee to the market fees
In accordance to [#24](https://github.com/TDex-network/tdex-specs/issues/24), the daemon must support adding a fixed fee to a market via the `UpdateMarketFee` RPC. This fee will be charged along with the percentage one when calculating the price of a trade.

### LINKED ISSUE #356: Add market balances to gRPC interface's MarketPrice RPC
At application level, we already added the balances to the reply of the GetMarketPrice method of the trade service.

In accordance with [#26](https://github.com/TDex-network/tdex-specs/issues/26), the balances should be returned also by the gRPC handler.

altafan:
This moves the operator.proto and wallet.proto directly into this project rather than having them in tdex-protobuf.
Along with this, a target `proto` is added to the Makefile in order to compile the Golang stubs.

This adds the fixed fee to the market, therefore the Operator service and the related CLI command too, change so that the original UpdateMarketFee is split into UpdateMarketPercentageFee and UpdateMarketFixedFee. 

NOTE: because now there are 2 ways of defining fees, the percentage fee can be zeroed, while before its lowest valid amount was 0.01%. The same applies for the fixed fee. Basically the operator has now the ability to zero the fees at his own will.

Balances are now returned into MarketPriceReply and amounts are uint64 instead of int64.

This closes #342.
This closes #353.
This closes #355.
This closes #356.

Please @tiero review this.


tiero (approved):

tiero (changes requested):
Can you add test for balanced preview formula?

PR #364, opened 2021-05-26
Add Support for webhooks

### LINKED ISSUE #360: Webhook capabilities
We need to allow external software to connect to the daemon to get notifications about events happening, such as a trade settled, balances too low, withdrawals etc..

altafan:
This adds the support for adding and removing webhooks registered to determined events (below those supported by now):
* Trade settled event
* Withdrawal from market event
* Account low balance (not supported yet)
* All (above) events

The flow is similar to the Github one, therefore the operator calls the AddWebhook RPC to register a webhook (an HTTP endpoint URL) for the preferred action. Once the event occurs, a POST request sent to the hook endpoint. The webhook can be secured, meaning that a secret can be specified which is used to sign the payloads of the future POST requests. The signature is sent as a bearer Json Web Token within the header `Authorization` of the request.

The info of the webhooks are stored in a secure lightweight key/value DB. This PR introduces a new package `securestore` that is an adaptation of the great effort produced by lnd team with macaroon's [RootKeyStorage](https://github.com/lightningnetwork/lnd/blob/master/macaroons/store.go).

NOTE: in theory, it should have been totally fine to use the [lnd/kvdb](https://github.com/lightningnetwork/lnd/blob/master/kvdb) package, but it's been required to copy down almost everything because there were issues with imports of the etcd subpkg which we don't want to import at all.

This closes #360.

Please @tiero review this.


api-spec/protobuf/operator.proto
```diff
@@ -236,6 +256,13 @@ enum TradeStatus {
   EXPIRED = 5;
 }
 
+enum ActionType {
+  TRADE_SETTLED = 0;
+  ACCOUNT_LOW_BALANCE = 1;
```
tiero, line 261: we want to distringuish between fee account and normal market account?

altafan: I thought it wasn't necessary... ideally, the payload of such request would include the account index to identify which account has low balance, along with other info like the account balance and, in case of a market, the asset pair.

In the past, it was the blockchain listener to signal through a log on the stdout when the fee/market's balance was too low. Now we need to put in place a proper logic; at the moment the only check the daemon performs on balances should be on TradePropose/MarketPrice, by checking if the balance is <= 0. This is the reason why you won't find the webhook invocation for the event AccountLowBalance.

internal/core/application/blockchain_listener.go
```diff
@@ -171,6 +172,29 @@ func (b *blockchainListener) listenToEventChannel() {
 				log.Warnf("trying to confirm or add unspents: %v", err)
 				break
 			}
+
+			// Invoke webhooks registered for action TradeSettled
+			if webhookManager != nil {
```
sekulicd, line 177: Method should not use external variable, webhookManager(maybe we should call all interfaces as Service, somewhere we use Manager somewhere Service) should be imported as a service(interface).

internal/core/application/webhook_handler.go
```diff
@@ -0,0 +1,297 @@
...
+		}
+		return nil, nil
+	})
+
+	return err
+}
+
+var webhookManager WebhookHandler
+
+func InitWebhookManager(store securestore.SecureStorage) (clean func(), err error) {
```
sekulicd, line 288: We should return instance of WebhookHandler, instead of initialising var which will be referenced throughout codebase.

sekulicd:
In general, Webhook is just fancy name for what is typical pub-sub(EventBus) pattern.
I think that core of our application should not be aware, nor coupled, to the implementation of Webhooks.
Ideally we should pass interface to our core, for  eg.
type EventBus interface {
    Publish(topic string, args ...interface{})
    Subscribe(topic string, fn interface{}) error 
    Unsubscribe(topic string, endpoint string) error
}
I envision that our core would just publish some 'Event'(data to topic) and in infrastructure we could implement logic(for eg. WebhookHandler could implement these methods) where we could invoke HTTP POST.
In Grpc handler we could invoke Subscribe or Unsubscribe which again would be implemented in infra where we would add or remove hook from DB.

altafan:
I partially agree with your vision of the EventBus interface: 
I do think that such service requires exposing methods to handle an internal database (create, lock, unlock, changepwd) so the operator keeps using the existing InitiWallet, UnlockWallet and ChangePassword RPCs to manage not only the "main" db but also the "side" ones like those for webhooks or the macaroons.

Maybe we can think about adding another service interface like
```
type PubSubService interface {
    DB() SecureStore
    EventBus() EventBus
}
``` 

Consider that when I started working on this PR I wanted to introduce domain/app events too, then had an offline chat with @tiero where we agreed upon postponing to a future PR so we can take the time to discuss the best approach and the proper design for core (domain/app) events.

So, for now, the methods of this service will be invoked where needed and we will make the necessary changes once we the core will emit events.

tiero:
> In general, Webhook is just fancy name for what is typical pub-sub(EventBus) pattern.
> I think that core of our application should not be aware, nor coupled, to the implementation of Webhooks.
> Ideally we should pass interface to our core/grpc handler, for eg.
> 
> ```
> type EventBus interface {
>     Publish(topic string, args ...interface{})
>     Subscribe(topic string, fn interface{}) error 
>     Unsubscribe(topic string, endpoint string) error
> }
> ```
> 
> I envision that our core would just publish some 'Event'(data to topic) and in infrastructure we could implement logic(for eg. WebhookHandler could implement these methods) where we could invoke HTTP POST.
> In Grpc handler we could invoke Subscribe or Unsubscribe which again would be implemented in infra where we would add or remove hook from DB.

I agree on the vision: this will allow to register many type of "triggers" with an internal core interface for subscribing and publishing events, that will be consumed either by webhooks, loggers, telemetry, notifications etc..

internal/core/application/operator_service.go
```diff
@@ -836,6 +834,29 @@ func (o *operatorService) WithdrawMarketFunds(
 		market.AccountIndex,
 	)
 
+	fmt.Println(txid)
```
tiero, line 837: remove

PR #370, opened 2021-06-10
Support auth macaroons

altafan:
This adds support for macaroon authentication over the gRPC daemon interface:

* `pkg/macaroons` has been added that is basically a copycat of [`lnd/macaroons`](https://github.com/lightningnetwork/lnd/blob/master/macaroons). This was required because importing that package means importing the whole `lnd` project, hence all its dependencies.
* A new interface [`Service`](https://github.com/altafan/tdex-daemon/blob/auth-macaroons/internal/interfaces/service.go) is defined in order to encapsulate all things relative to a specific interface whether it is gRPC, REST, or whatever. All stuff that in `main.go` related to to gRPC, therefore, has been moved to the gRPC implementation of `Service`.
This implementation makes sure to create the default macaroon files `admin.macaroon`, `readonly.macaroon`, `market.macaroons` and `price.macaroons` if they are not found in the datadir when Initializing or Unlocking the internal wallet.
It's important to note that macaroon auth is handled at Interface level, and the dedicated macaroon service needs the password(s) sent into the Init/unlock/changepwd request to manage its internal encrypted store accordingly. For this reason, the `application.WalletService` now exposes a [method](https://github.com/altafan/tdex-daemon/blob/auth-macaroons/internal/core/application/wallet_service.go#L65) used to retrieve the channel where it sends the password(s) to for successfully served requests. This way, "reverse" communication from a lower to an upper level is accomplished.
Macaroons auth requires a not insecure connection, therefore, when enabled, the daemon automatically creates a TLS key and a self-signed certificate to encrypt the communication channel. Macaroons auth, when enabled, "protects" the Operator service RPCs and all those of the Wallet one but GenSeed, Init, Unlock and ChangePassword which are left "unprotected".
TLS encryption for the Trade interface has been left as it already was. These changes do not affect what previously done, except for the fact that the old `sslcert` script has been renamed `tlscert`,as the env var prefix is changed from SSL to TLS  for a more accurate nomenclature.
Last but not least, a new `TDEX_NO_MACAROONS` env var has been added to disable macaroons auth and relative TLS encryption (mainly for dev purposes).
* The end-to-end test has been fixed so that it runs with a fully insecure, authenticated connection. Would be nice to have also one that uses TLS and macaroons, but I didn't add it yet because I'm concerned about the increasing waiting time to make it run. What do you think about it?

This takes place of #300 which should be closed.

Please @tiero @sekulicd review this.


tiero:
Restore the SSL env variables, those are neeeed for server side certificates for the trader interface.

Also, no need for a NO TLS var, the NO MACAROONS should be used to signal to use auth (and TLS self signed along with it)

The make cert target in makefile was used for development purposes to test the traderr interface TLS

altafan:
Done. Check the updated PR description.

internal/interfaces/grpc/service.go
```diff
@@ -0,0 +1,469 @@
...
+	// Location is used as the macaroon's location hint. This is not verified as
+	// part of the macaroons itself. Check the doc for more info:
+	// https://github.com/go-macaroon/macaroon#func-macaroon-location.
+	Location = "tdexd"
+	// DbFile is the name of the macaroon database file.
+	DBFile = "macaroons.db"
+	// AdminMacaroonFile is the name of the admin macaroon.
+	AdminMacaroonFile = "admin.macaroon"
+	// AdminMacaroonFile is the name of the read-only macaroon.
+	ReadOnlyMacaroonFile = "readonly.macaroon"
```
sekulicd, line 53: fix comments

sekulicd:
> It's important to note that macaroon auth is handled at Interface level, and the dedicated macaroon service needs the password(s) sent into the Init/unlock/changepwd request to manage its internal encrypted store accordingly. For this reason, the application.WalletService now exposes a method used to retrieve the channel where it sends the password(s) to for successfully served requests. This way, "reverse" communication from a lower to an upper level is accomplished.

What is the reasoning behind moving auth from interceptor to interface layer?

altafan:
> What is the reasoning behind moving auth from interceptor to interface layer?

The mcaroon auth cheks are still performed within the gRPC interceptor, which is part of the interface layer, more precisely it's part of the gRPC interface. Just like you did in the previous PR :)

We refer to interface layer as that one grouping all the different ways to communicate with the daemon core application from the outside world, so for example gRPC, REST, JSON-RPC etc. 

NOTE: if, for any reason, in the future we'll need to add a new interface, then macaroon auth will be required to be implemented also by that one, like for example a middleware checking the mac in the request header.
The rationale is that every interface has its very specific way to pass auth data from the client to the server: in gRPC the macaroon needs to be passed within request metadata, in REST, instead, as already said, within the request headers, etc, therefore, IMO it's totally fine that authentication requires to be implemented for every interface.

config/config.go
```diff
@@ -20,16 +19,16 @@ import (
...
 	// OperatorListeningPortKey is the port where the gRPC Operator interface will listen on
 	OperatorListeningPortKey = "OPERATOR_LISTENING_PORT"
 	// ExplorerEndpointKey is the endpoint where the Electrs (for Liquid) REST API is listening
 	ExplorerEndpointKey = "EXPLORER_ENDPOINT"
 	// ExplorerRequestTimeoutKey are the milliseconds to wait for HTTP responses before timeouts
 	ExplorerRequestTimeoutKey = "EXPLORER_REQUEST_TIMEOUT"
-	// DataDirPathKey is the local data directory to store the internal state of daemon
-	DataDirPathKey = "DATA_DIR_PATH"
+	// DatadirKey is the local data directory to store the internal state of daemon
+	DatadirKey = "DATADIR"
```
tiero, line 31: this likely breaking change, ie. docker container, documentation etc..

config/config.go
```diff
@@ -46,10 +45,10 @@ const (
 	TradeExpiryTimeKey = "TRADE_EXPIRY_TIME"
 	// PriceSlippageKey is the percentage of the slipage for accepting trades compared to current spot price
 	PriceSlippageKey = "PRICE_SLIPPAGE"
-	// SSLCertPathKey is the path to the SSL certificate
-	SSLCertPathKey = "SSL_CERT"
-	// SSLKeyPathKey is the path to the SSL private key
-	SSLKeyPathKey = "SSL_KEY"
+	// TradeTLSKeyKey is the path of the the TLS key for the Trade interface
+	TradeTLSKeyKey = "TRADE_TLS_KEY"
```
tiero, line 49: Also this breaking chnage. lets propose and do in a separate PR if any

config/config.go
```diff
@@ -89,29 +99,25 @@ func init() {
...
-	if err := initDataDir(); err != nil {
-		log.WithError(err).Panic("error while init data dir")
+	if err := initDatadir(); err != nil {
+		log.WithError(err).Panic("error while creating datadir")
 	}
 
 	vip.Set(MnemonicKey, "")
 }
 
-func makeDirectoryIfNotExists(path string) error {
```
tiero, line 108: wy removing?

cmd/tdex/main.go
```diff
@@ -195,9 +201,57 @@ func getClientConn() (*grpc.ClientConn, error) {
...
+		certPath := filepath.Join(certDatadir, tlsCertFile)
+
+		tlsCreds, err := credentials.NewClientTLSFromFile(certPath, "")
+		if err != nil {
+			return nil, fmt.Errorf("could not read TLS certificate:  %s", err)
+		}
+		opts = append(opts, grpc.WithTransportCredentials(tlsCreds))
+
+		// Load macaroons and add credentials to dialer
+		if !skipMacaroon {
```
tiero, line 229: whats difference between nomacaroons and skip

internal/interfaces/grpc/macaroon_utils.go
```diff
@@ -0,0 +1,109 @@
+package grpcinterface
```
tiero, line 1: rename just macaroons

internal/interfaces/service.go
```diff
@@ -0,0 +1,11 @@
+package interfaces
+
+// Service interface defines the methods that every kind of interface, whether
+// gRPC, REST, or whatever must be complaint with.
```
tiero, line 4:
```suggestion
// gRPC, REST, or whatever must be comply with.
```

cmd/tdex/main.go
altafan: the trade interface isn't dependent from the macaroon authentication. `skipMacaroons` is basically used by serveMux to distinguish between the trade and the operator interface. For the former macaroon is always disabled, while for the latter it depends on the configuration.

tiero (approved):

PR #374, opened 2021-06-15
Add extraIP/extraDomain to Operator TLS cert & refactor macaroon permissons/files

### LINKED ISSUE #372: Markets do not shows fixed_fee

### LINKED ISSUE #373: Cannot connect to remote dockerized daemon with macaroon auth + TLS encryption
I tried to spin up a dockerized daemon using the image `ghcr.io/tdex-network/tdexd:v0.4.0-rc.1` on a remote machine with macaroons auth + TLS encryption over the Operator interface.

I have no problems using the commands of the remote host's CLI after configuring the macaroons and TLS certificate paths.

I do have problems, though, while trying to use my local key connected to the remote host:

1. The CLI always expected to find an `admin.macaroon` in the `macaroons_path` but I'm not able to _scp_ copy that file from the remote host to the local one. An error `permission denied` occurs because that file is created with permissions `600` and requires `sudo` to be copied. `admin.macaroon` is the only macaroon that enables the user to use all the commands of the CLI. It could be possible to use `readonly.macaroon` (renamed admin.macaroon in my machine) to have at least access to the "read" commands, but as long as we don't have a bakery command to create new macaroons from those existing, we might need to loosen the permissions of `admin.macaroon` so that it can be exported (perm `644` is fine).
2. I cannot make use of the `cert.pem` TLS certificate created by the daemon from my local machine: I get the error
   ```
   [tdex] rpc error: code = Unavailable desc = connection error: desc = "transport: authentication handshake failed: x509: certificate is valid for d403831341a0, localhost, unix, unixpacket, not ********
   ```
   The reason is that the daemon is running in a docker container and unless the container is running in the _host_ network, the daemon won't be able to access the [hostname](../blob/master/internal/interfaces/grpc/service.go#L338) of the machine it is running on. In this case, we need to add some new env vars to communicate to the daemon an extra IP address and/or dns name to add to those validated by the self-signed certificate.

altafan:
This adds 2 new env vars TDEX_OPERATOR_EXTRA_IP and TDEX_OPERATOR_EXTRA_DOMAIN in order to add an IP address or a DNS domain to the Operator TLS certificate. This becomes useful when running dockerized daemons, or serving them as TOR hidden services.

The macaroons permissions have been modified in order to have more than just the operator and trade entities. Other entities have been added: _market_, _price_, _wallet_, _webhook_ and _unlocker_. This way, defining permissions becomes smoother and instead of defining permissions for each RPC method, it's possible to define actions for entities.
With this, 2 new macaroons `wallet.macaroon` and `webhook.macaroon` in addition to the others.

For what concerns the CLI, the hardcoded filenames for the macaroons file and the TLS certificate have been removed and now it's up to the operator to specify the name of those files within `macaroons_path` and `tls_cert_path` respectively.

Info about the market balances have been added to the payload for TRADE_SETTLED event-registered webhooks.

BONUS: The reply of the Trade.Markets RPC have been fixed returning also the fixed fee info.

This closes #373 
This closes #372.

Please @tiero review this.

### DIFF
--- a/internal/interfaces/grpc/service.go
+++ b/internal/interfaces/grpc/service.go
@@ -59,6 +59,9 @@ const (
 	// PriceMacaroonFile is the name of the macaroon allowing to update only the
 	// prices of markets.
 	PriceMacaroonFile = "price.macaroon"
+	// WebhookMacaroonFile is the name of the macaroon allowing to add, remove or
+	// list webhooks.
+	WebhookMacaroonFile = "webhook.macaroon"
 )
 
 var (
@@ -76,10 +79,12 @@ type service struct {
 type ServiceOpts struct {
 	NoMacaroons bool
 
-	Datadir           string
-	DBLocation        string
-	TLSLocation       string
-	MacaroonsLocation string
+	Datadir             string
+	DBLocation          string
+	TLSLocation         string
+	MacaroonsLocation   string
+	OperatorExtraIP     string
+	OperatorExtraDomain string
 
 	WalletSvc   application.WalletService
 	OperatorSvc application.OperatorService
@@ -117,6 +122,10 @@ func (o ServiceOpts) validate() error {
 				OperatorTLSKeyFile, OperatorTLSCertFile, tlsDir,
 			)
 		}
+
+		if o.OperatorExtraIP != "" && net.ParseIP(o.OperatorExtraIP) == nil {
+			return fmt.Errorf("invalid operator extra ip %s", o.OperatorExtraIP)
+		}
 	}
 	if o.WalletSvc == nil {
 		return fmt.Errorf("wallet app service must not be null")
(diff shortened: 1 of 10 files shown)


internal/interfaces/grpc/service.go
```diff
@@ -313,6 +338,10 @@ func generateOperatorTLSKeyCert(datadir string) error {
 	// Collect the host's IP addresses, including loopback, in a slice.
 	ipAddresses := []net.IP{net.ParseIP("127.0.0.1"), net.ParseIP("::1")}
 
+	if extraIP != "" {
+		ipAddresses = append(ipAddresses, net.ParseIP(extraIP))
```
tiero, line 342: is is not a valid IP it will return nil, is this correctly?

PR #393, opened 2021-06-28
Move preview to domain level

### LINKED ISSUE #368: Refactor preview
At the moment it is hard to unit test previews for markets. All the logic relative to what kind of preview to make based on the market strategy etc resides in the application layer, but we do want to have only component tests at that level.

My proposal is to add a `Preview()` method (args TBD) to the market domain, where to basically move all the logic currently living in trade_service.go, with the purpose of streamlining the code of the app layer to just a function call.

It is true that such method would not change the internal state of the domain so technically should not be part of it, but it's also true that a preview is strictly related to a market (the only "external" data are its balances).
The PROs, to me, are evident:
* Changes at domain level, like for example adding a strategy, would not require changes at app level, like creating an internal function for calculating previews for that specific strategy.
* we would be able to write unit test of previews for all kind of strategies, making it easier to find&fix bugs such as #366 in the future.

@tiero @sekulicd @louisinger do you have any feedback?

altafan:
This moves the whole price preview logic to the domain level, with the purpose to abstract the handling of a preview based on the market strategy at with a simple `market.Preview()` invocation at application level.

With this, a new `PluggableStrategy` data type is defined along with its method to make it compliant with the `mm.MakingFormula` interface.

The methods InGivenOut and OutGivenIn of the interface just mentioned have been slightly changed so that they accept opts of any type by expecting `interface{}`. It is then responsibility of the implementation to make sure that the opts are of the correct type by checking whether the interface casting succeeds or not.

Within the _balanced_ AMM formulas, the usage of the ChargeFeesOnThe WayIn has been fixed: before this that bool was basically a charge/discount fee flag. Now, the fees are **always charged**, and the bool states whether to apply fees on the given amount or the calculated one. The daemon charges fees always on the input amount.

This closes #368.

Please @tiero, review this.

### DIFF
--- a/internal/core/domain/market_service.go
+++ b/internal/core/domain/market_service.go
@@ -275,18 +275,14 @@ func (m *Market) formula(
 	formula := m.getStrategySafe().Formula()
 	if isBuy {
 		if isBaseAsset {
-			// fmt.Println("InGivenOut")
 			return formula.InGivenOut
 		}
-		// fmt.Println("OutGivenIn")
 		return formula.OutGivenIn
 	}
 
 	if isBaseAsset {
-		// fmt.Println("OutGivenIn")
 		return formula.OutGivenIn
 	}
-	// fmt.Println("InGivenOut")
 	return formula.InGivenOut
 }
 
--- a/internal/core/domain/market_service.go
+++ b/internal/core/domain/market_service.go
@@ -238,7 +238,12 @@ func (m *Market) Preview(
 	}
 
 	formula := m.formula(isBaseAsset, isBuy)
-	args := m.formulaOpts(baseBalance, quoteBalance, isBaseAsset, isBuy)
+	var args interface{}
+	if m.IsStrategyPluggable() {
+		args = m.formulaOptsForPluggable(baseBalance, quoteBalance, isBaseAsset, isBuy)
+	} else {
+		args = m.formulaOptsForBalanced(baseBalance, quoteBalance, isBaseAsset, isBuy)
+	}
 
 	price, err := m.priceForStrategy(baseBalance, quoteBalance)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


internal/core/domain/market_service.go
```diff
@@ -217,6 +219,176 @@ func (m *Market) ChangeQuotePrice(price decimal.Decimal) error {
...
+func (m *Market) Preview(
+	baseBalance, quoteBalance, amount uint64,
+	isBaseAsset, isBuy bool,
+) (*PreviewInfo, error) {
+	if !m.IsTradable() {
+		return nil, ErrMarketIsClosed
+	}
+
+	if isBaseAsset {
+		if amount < uint64(m.FixedFee.BaseFee) {
```
tiero, line 231: I would argue that also equal to fixed fee, is too low.

Also we do not take into consideration that if percentage fee is greater than zero, wuold not be enough.

internal/core/domain/market_service.go
```diff
@@ -217,6 +219,176 @@ func (m *Market) ChangeQuotePrice(price decimal.Decimal) error {
...
+		return formula.OutGivenIn
+	}
+
+	if isBaseAsset {
+		return formula.OutGivenIn
+	}
+	return formula.InGivenOut
+}
+
+func (m *Market) formulaOpts(
```
tiero, line 289: I would rather uplift the branch (if strategy or not) and create a dedicated method like `formulaOptsForPluggable`, `formulaOptsForBalanced` mostly beacuse we can this way add more formulas later, otherwise this methos will hide all the logic.

internal/core/domain/market_service.go
altafan: > Also we do not take into consideration that if percentage fee is greater than zero, wuold not be enough.

The strategy takes care of charging the percentage fee, and, in case the fee must be subtracted, we check that after appliying the fees the amount is still positive.

It's not possible to say if the amount is "enough" for perc fees but doing the math "empirically" and checking the calculated amount is > 0.

> I would argue that also equal to fixed fee, is too low.

My though was: if the amount is at least equal to the respective fixed fee then it's allright because the perc fee will be applied. I'll add the equality check if this way of thinking does not convince you.

tiero: receiving 0 output, although doable, not that much meaningful for the user

PR #396, opened 2021-07-01
Split wallet and walletunlocker services

### LINKED ISSUE #390: Add isReady RPC to let other services know when the daemon is avaiable to be ie. unlocked or TLS certificate is created

altafan:
This adds a new proto service `WalletUnlocker` that isolates some of the methods originally defined in the `Wallet` service:

```
service WalletUnlocker {
  rpc GenSeed(GenSeedRequest) returns (GenSeedReply);
  rpc InitWallet(InitWalletRequest) returns (stream InitWalletReply);
  rpc UnlockWallet(UnlockWalletRequest) returns (UnlockWalletReply);
  rpc ChangePassword(ChangePasswordRequest) returns (ChangePasswordReply);
  rpc IsReady(IsReadyRequest) returns (IsReadyReply);
}
```

The handler for this service is still registered on the so-called Operator interface running by default on `:9000`. Starting from this, the gRPC Service implementation registers only the handler for WalletUnlocker and until the wallet is initialized from scratch no other handlers and interfaces will be available.
Before this, in case the daemon has been restarted the restoration of the UTXO set was locking, causing the Operator and Trade interfaces to not be available until it completed.
Now, on the contrary, the UTXO restoration is handled by a goroutine and the WalletUnlocker app service provides a method to source from a channel when it has completed the initialization.
When the app service signals that it's ready, the parent gRPC service gets notified and registers the operator and wallet handlers on the Operator interface and enables the Trade interface.

BONUS: it is now possible to unlock the wallet while it's still syncing the UTXO set. While this gives no advantage to the operator because other handlers but the WalletUnlocker are not enabled, this feature will turn useful when adding the auto-unlocker service.

This closes #390.

Please @tiero review this.


tiero:
solve conflcits @altafan

internal/core/application/walletunlocker_service.go
```diff
@@ -0,0 +1,950 @@
...
+func (w *walletUnlockerService) GenSeed(ctx context.Context) ([]string, error) {
+	mnemonic, err := wallet.NewMnemonic(wallet.NewMnemonicOpts{EntropySize: 256})
+	if err != nil {
+		return nil, err
+	}
+	return mnemonic, nil
+}
+
+func (w *walletUnlockerService) IsReady(_ context.Context) bool {
+	return w.isInitialized()
```
tiero, line 145: is it isInitiazed used to know if the seed is there? wouldnt be better adding another state like `isRestored`

so we check if `isInitialized` && `isRestored`

altafan: `isInitialized` expresses the condition of a wallet that has been created/loaded from db and its UTXO set is fully synced.
The service knows if a wallet has been created by attempting to get the vault form the db and checking if it's nil eventually.

sekulicd:
@altafan what was the motivation for separating WalletUnlocker? why do we need to disable other RPCs before init?
wasnt returning error that some RPC is not available enough? IMHO disabling RPCs till wallet is initialized adds additional complexity, was it necessery?

tiero:
> @altafan what was the motivation for separating WalletUnlocker? why do we need to disable other RPCs before init?

The purpose is to be able to unlock BEFORE the restore of unspents is finished.

IMO better to refer to `restored`, rather than `initialized` which means adding the mnemonic to vault.

> wasnt returning error that some RPC is not available enough? 

We can register all the services, but this would makes harde to maintain no? because you need to add this check (isRestored) in all other RPCs that should be called before restoration been complete, this way we only expose the one we need?

altafan:
> We can register all the services, but this would makes harde to maintain no? because you need to add this check (isRestored) in all other RPCs that should be called before restoration been complete, this way we only expose the one we need?

Also, the problem with this approach would have been having all app services "depending" on the wallet unlocker one, while we do want them to be standalone.

With the approach proposed, app services continue to be standalone, ignoring other services as designed. The "problem" is solved at interface layer  instead by preventing interacting with some app services based on its internal status.

tiero:
So final review/comment @sekulicd

sekulicd:
ACK

tiero:
I do think would be more semantically correct split these RPC in even more proto services: Init and GenSeed in a `WalletInitializer` and ChnagePassword and Unlock in `WalletUnlocker` and isReady is a `HealtChecker` service.

altafan:
I'm not so conviced about this kind of service fragmentation. Why should we separate Init from Unlock and ChangePassphrase? They all manage the main password to lock/unlock the internal wallet, while GenSeed is just an util function that could be fairly bypassed by the user if he wants to use his personal entropy and rng to be 100% sure of the randomness of his seed.
For sure I think we need an HealthChecker service to ping the daemon and check if it's online, but IsReady is not properly related to this. It gives information about the internal state of the daemon so it's more WalletUnlocker context than HealtChecker.

Furthermore, all these new changes would require breaking changes to the macaroons permissions, so that identities reflect the services. Any daemon outthere using macaroons auth would require the deletion and recreation of the macaroons file after updating to this new version.

tiero:
Just semantically WalletUnlocker doesent seem right for the others RPC.

>  all these new changes would require breaking changes to the macaroons permissions

Why so? WalletUnlocker has been split in this PR anyway. Macaroon wise nothing should change, we add in the same scope different services

altafan:
> Just semantically WalletUnlocker doesent seem right for the others RPC.

If you prefer we can change the name of the service but I do think it has much more sense keeping these RPCs alltogether rather than splitting just because of a service name.
They're all relative to the pre-unlock and unlock phases. Introducing services just because a name is not 100& suitable is not a good reason IMHO.

I don't see any advantage in introducing more services. On the contrary, this would increase the complexity of the code and therefore its readability and maintainability.

altafan:
> Why so? WalletUnlocker has been split in this PR anyway. Macaroon wise nothing should change, we add in the same scope different services

WalletUnlocker was already existing as a macaroon identity to group all these "unprotected" RPCs (GenSeed, Init, Unlock, ChangePassphrase, and now also IsReady) when they were unified to the RPCs of what is now called Wallet service.

PR #397, opened 2021-07-16
New unlocker service, IsReady RPC response type and default cert/macaroons for CLI

### LINKED ISSUE #395: Auto Unlocker
now, in the occasion the daemon goes offline and get restarted, the trading operation will not be available unless the daemon operator unlock (either via RPC from a remote or ssh-in and using the embedded CLI)

therefore we want to have a dedicated binary, that runs on the side on the same machine and can be configured to auto unlock after the daemon is ready and stores the password in many secrets providers (we need to integrate), either AWS KMS, Kubernetes Secrets, Hashicorp vault etc..

to do this we need to make some changes to the daemon (ie. allowing the UnlockWallet RPC to be reachable ASAP, expose a new isReady RPC that other micro-services or this new auto-unlocker binary for example can call to know when can be unlocked etc..)

altafan:
## IsReady RPC

This slightly modifies the reply of the IsReady RPC of the WalletUnlocker service so that it returns more info about the status of the daemon rather than just a simple bool. [Here](https://github.com/altafan/tdex-daemon/blob/auto-unlock/api-spec/protobuf/walletunlocker.proto#L117) you can find the definition of the new IsReadyRepy message.

## Unlockerd

This new change, along with other already committed, paves the way for the new `unlockerd` auto-unlock service. This new binary sources the wallet password and possibly the TLS certificate from some *provider*.

A provider must be compliant with the following interface in order for the main service to be able to source the password (and possibly the TLS certificate) to unlock its internal wallet:

```
type provider interface {
   Password() ([]byte, error)
   TLSCertificate ([]byte, error)
}
```

At the moment, the only *provider* supported is `file`, therefore the user must specify a password filepath from where to read the password to unlock the wallet. By default, this provider uses the default daemon datadir to know if it has to load a TLS certificate. Alternatively, It is possible to specify the exact path to the certificate file with the apposite flag.

Other providers will be added with future PRs.

## Tdex CLI

Bonus: This friendly behaviour of the unlocker, which sources the certificate from the default daemon datadir if not specified otherwise, has been adopted also by the CLI (also for sourcing the admin.macaroon file).

This closes #395.

Please @tiero review this.


tiero:
Can you add to Dockerfile and GoReleaser?

cmd/unlockerd/main.go
```diff
@@ -0,0 +1,260 @@
...
+	return k
+}
+
+type provider interface {
+	Password() ([]byte, error)
+	TLSCertificate() ([]byte, error)
+}
+
+const (
+	rpcAddressKey = "rpc-address"
```
tiero, line 42: `rpcserver` to be consistent to tdex cli?

cmd/unlockerd/main.go
```diff
@@ -0,0 +1,260 @@
...
+	return true
+}
+
+func cleanAndExpandPath(path string) string {
+	if path == "" {
+		return ""
+	}
+
+	// Expand initial ~ to OS specific home directory.
+	if strings.HasPrefix(path, "~") {
```
tiero, line 209: why we do not use btcutil HomeDir method? which is the same used by the daemon to create his datadir?

cmd/unlockerd/fileprovider.go
```diff
@@ -0,0 +1,65 @@
...
+	tlsCertPath := viper.GetString(tlsCertKey)
+	if tlsCertPath == "" {
+		return nil, fmt.Errorf("invalid flag: %s must not be null", tlsCertKey)
+	}
+
+	return &fileProvider{pwdPath, tlsCertPath}, nil
+}
+
+func (fp *fileProvider) Password() ([]byte, error) {
+	return ioutil.ReadFile(fp.pwdPath)
```
tiero, line 57: ok maybe worht adding a README where we document the file format. it looks like we assume the file will just have the password there in it and we read all until EOF?

tiero (changes requested):
Using dash is a bit inconsistent with the underscore used  by the tdex CLI. worth keep in sync

cmd/unlockerd/fileprovider.go
altafan: exactly, the file should contain ONLY the password and nothing else. Maybe we could add a section dedicated to build&use the unlocker binary to the main README just like we did for the CLI.
Then, its detailed usage could be documented into dev-portal?

cmd/unlockerd/main.go
altafan: I actually copied the `cleanAndExpandPath` method from the [cli](../blob/master/cmd/tdex/main.go#L258).

tiero (approved):

PR #402, opened 2021-07-26
[CLI] Auto create state file if not existing

### LINKED ISSUE #341: [tdex] auto create a state.json with defaults in the datadir. config init command should be able to overwrite

altafan:
This makes the previously mandatory step `config init [--network, ...]` optional.

Now, when the `getState` function is in charge of creating a state.json file with all default values in case it doesn't exist instead of returning an error that mentions to run `config init`.

This closes #341.

Please @tiero review this.

### DIFF
--- a/cmd/tdex/config.go
+++ b/cmd/tdex/config.go
@@ -19,26 +19,31 @@ const (
 )
 
 var (
-	daemonDatadir        = btcutil.AppDataDir("tdex-daemon", false)
+	daemonDatadir = btcutil.AppDataDir("tdex-daemon", false)
+
+	defaultNetwork       = network.Liquid.Name
+	defaultExplorer      = "https://blockstream.info/liquid/api"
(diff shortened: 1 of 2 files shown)


cmd/tdex/config.go
```diff
@@ -19,26 +19,31 @@ const (
 )
 
 var (
-	daemonDatadir        = btcutil.AppDataDir("tdex-daemon", false)
+	daemonDatadir = btcutil.AppDataDir("tdex-daemon", false)
+
+	defaultNetwork       = network.Liquid.Name
+	defaultExplorer      = "https://blockstream.info/liquid/api"
+	defaultRPCServer     = "localhost:9000"
+	defaultMacaroonsAuth = false
```
tiero, line 27:
```suggestion
	defaultNoMacaroonsAuth = false
```

tiero: @altafan

PR #403, opened 2021-07-28
Watch outpoints instead of transaction ids when a proposal is accepted

### LINKED ISSUE #369: Support detecting swaps settlement made with non-segwit wallets
After the first trade, the deamon quote asset balance becomes zero.
Steps I followed:
- initialize the deamon with 100000000 units of the minted coin
- made a trade in order to buy 45000000 units
- call the `Balances` gRPC rpc
Here the response:
```json
{
  "balances": [
    {
      "balance": {
        "base_amount": "100000000",
        "quote_amount": "0"
      },
      "fee": {
        "basis_point": "25",
        "fixed": null
      }
    }
  ]
}
```
Obviously all the subsequent trade propose calls are failing. I also checked the minted coin balance of the client and it actually contains only 45000000 units.

### LINKED ISSUE #389: Add settlement timestamp and Date Time to TRADE_SETTLED event

altafan:
## Overwiew

Before this, after a proposal was accepted, the daemon would have started watching for the id extracted from the related partial trasaction. This is fine only for "full" native-segwit transactions where the hash of the tx doesn't change before and after it being singed. Unfortunately, as well known, this doesn't hold in case the tx included wrapped-segwit or legacy inputs.

For this reason, this adds the required changes to make the daemon watching for single outpoints, or the inputs of the spending tx, and now acquires the txid from the blockchain instead of applying the logic described above.

## Explorer

It was necessary to add a new [method](https://github.com/altafan/tdex-daemon/blob/observe-utxo/pkg/explorer/explorer.go#L67) to the explorer interface (and related implementations) to get the status of some utxo.
This status is represented by an [interface](https://github.com/altafan/tdex-daemon/blob/observe-utxo/pkg/explorer/explorer.go#L29) and the same has been done for the returned value of the already existing  [GetTransactionStatus](https://github.com/altafan/tdex-daemon/blob/observe-utxo/pkg/explorer/explorer.go#L47) for consistency and clean programming.

## Crawler

This package has been refactored a bit:
* the observable handler type and its functions have been exported to make it possible to unit-test all kind of observable and checking the event/error they eventually emit.
* AddressObservable has been decoupled from the core's internal Vault domain. The original AccountIndex field has been replace by a general `ExtraData` interface returned in the event emitted.
* A new type `OutpointsObservable` has been added useful to observe the status of a group of outpoints, all used as inputs of the same transacton. This observable emits events in case the outpoints have not yet been spent, or if they've been spent and the tx is in mempool, or if they've been spent and the tx is in blockchain.

## Trade Service

This service now instructs the bc listener to watch for outpoints instead of transactions.

## Blockchain Listener

Added support for adding/removing OutpointsObservables and to handle events they possibly emit: in case the outpoints have been spent but the tx is not yet confirmed, its outs are added as new unconfirmed utxos to the repo, while its ins are marked as SPENT; when the tx is confirmed the new utxos get confirmed as well in the repo. Only in case the spending tx is confirmed the trade gets settled and the topic pushed in the pubsub service.

## Bonus
* Settlement timestamp and date added to the TRADE_SETTLED event payload
* Updated jwt lib to the new path

This closes #389.
This closes #369.

Please @tiero review this.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,8 +13,8 @@ require (
 	github.com/btcsuite/btcutil v1.0.3-0.20201208143702-a53e38424cce
 	github.com/desertbit/timer v0.0.0-20180107155436-c41aec40b27f // indirect
 	github.com/dgraph-io/badger/v2 v2.2007.2
-	github.com/dgrijalva/jwt-go v3.2.0+incompatible
 	github.com/gogo/protobuf v1.2.1
+	github.com/golang-jwt/jwt v3.2.1+incompatible
 	github.com/golang/protobuf v1.4.3
 	github.com/google/uuid v1.1.2
 	github.com/grpc-ecosystem/go-grpc-middleware v1.0.1-0.20190118093823-f849b5445de4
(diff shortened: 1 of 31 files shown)


tiero:
Solve conflicts @altafan

tiero:
Seems not working

```
DEBU[46884] unlocked 2 unspents
INFO[46884] trade with id 6e1bc520-fdcc-4aa4-a960-e48eb5587251 expired
```

Log 

[log.txt](https://github.com/TDex-network/tdex-daemon/files/6937904/log.txt)

tiero (approved):

PR #412, opened 2021-08-20
List deposits list withdrawals

### LINKED ISSUE #391: Add ListDeposits and ListWithdrawals RPCs
Like for `ListTrades` RPC, we should let the operator to download the his own deposits and the withdrawals. Would be better if the user could specify which account is interested (fee account, wallet account or a specific markets).

Since it can have lot of entries, better to allow the RPC to have pagination-like feature.

This means a new domain/storage should be added to keep track of these operations (new deposit after a claim, new withrawal after withdrawal RPC and so on).

### LINKED ISSUE #404: Check account index in DropMarket
The DropMarket RPC should enforce that the providd accountIndex is at least >= `domain.MarketAccountStart`.

sekulicd:
This adds implementation for ListDeposits and ListWithdrawals RPCs.

This closes #391 
This closes #404 


tiero (commented):
Why not having a badger db/repo for deposits  and one for withdrawals?

Conceptually its like the Trade domain/repo

sekulicd:
> Why not having a badger db/repo for deposits and one for withdrawals?
> 
> Conceptually its like the Trade domain/repo

I was thinking that if we use deposits and withdrawals only in these scenarios, to simplify it and not having separate domains for each and also leave room if similar requirement occurs in future.

api-spec/protobuf/operator.proto
```diff
@@ -362,3 +369,34 @@ message WebhookInfo {
...
+message ListWithdrawalsRequest{
+  int64 account_index = 1;
+  Page page = 2;
+}
+message ListWithdrawalsReply{
+  int64 account_index = 1;
+  repeated Withdrawal withdrawals = 2;
+}
+
+message Withdrawal {
```
altafan, line 392: this message should include also the txid of the withdrawal tx

api-spec/protobuf/operator.proto
```diff
@@ -362,3 +369,34 @@ message WebhookInfo {
...
 }
+
+
+message ListDepositsRequest{
+  int64 account_index = 1;
+  Page page = 2;
+}
+message ListDepositsReply{
+  int64 account_index = 1;
+  repeated TxOutpoint deposits = 2;
```
altafan, line 380: would be better to be:
```suggestion
  repeated UtxoInfo deposits = 2;
```

since it that kind of message gives also info about the asset and amount deposited.

internal/core/domain/stats_model.go
```diff
@@ -0,0 +1,31 @@
+package domain
+
+// Withdrawal is used to follow funds withdrawal statistics
+type Withdrawal struct {
+	ID              uint `badgerhold:"key"`
```
altafan, line 5: domain MUST not have any reference to any kind of concrete db whether it is badger, postgres or whatever. Please remove. Also, a withdrawal is already uniquely identified by the tx id included in blockchain, therefore there's no need to have another one.

internal/core/domain/stats_model.go
```diff
@@ -0,0 +1,31 @@
...
+	AccountIndex    int
+	BaseAmount      uint64
+	QuoteAmount     uint64
+	MillisatPerByte int64
+	Address         string
+}
+
+// Deposit is used to follow deposit statistics made by operator
+type Deposit struct {
```
altafan, line 14: this should define other 2 fields (asset and amount) according to the previous suggestion on proto definition.

internal/core/domain/stats_repository.go
```diff
@@ -0,0 +1,23 @@
+package domain
+
+import "context"
+
+type Page struct {
+	Number int
+	Size   int
+}
+
+type StatsRepository interface {
```
altafan, line 10: I think it's better to have 2 distinct repos for withdrawals and deposits. Considering these as 2 different tables (for relational DBs) or buckets (for key-value DBs), or in general DB files, IMO it would be better to stick with the concept "one repo - one db file". This way, in a certain way we are enforcing at the domain level that things in the concrete DB must be separated in order to grant exlusive lock on only a single file (wheter it contains a table or a bucket).

This would make the operator able to do a withdrawal while depositing some funds. If withdrawals and deposits go to the same file that would not be possible so to speak.

internal/infrastructure/storage/db/test/stats_repository_impl_test.go
```diff
@@ -0,0 +1,118 @@
+package db
```
altafan, line 1: Please stick with the code structure of other test files.

altafan (changes requested):

internal/core/domain/stats_repository.go
tiero: agreed @sekulicd

api-spec/protobuf/operator.proto
sekulicd: Done

internal/core/domain/stats_model.go
sekulicd: Done

internal/core/domain/stats_model.go
sekulicd: Done

internal/core/domain/stats_repository.go
sekulicd: Done

internal/infrastructure/storage/db/test/stats_repository_impl_test.go
sekulicd: Done

internal/infrastructure/storage/db/test/deposit_repository_impl_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+			t.Run("testAddAndListDeposits", func(t *testing.T) {
+				t.Parallel()
+				testAddAndListDeposits(t, repo)
+			})
+		})
+	}
+}
+
+func testAddAndListDeposits(t *testing.T, repo depositRepository) {
```
altafan, line 33: we should check that attempting to add the same deposit/withdrawal more than once does not create copies into the repo impl (redundancy).

internal/infrastructure/storage/db/test/withdrawal_repository_impl_test.go
```diff
@@ -0,0 +1,108 @@
...
+			t.Run("testAddAndListWithdrawals", func(t *testing.T) {
+				t.Parallel()
+				testAddAndListWithdrawals(t, repo)
+			})
+		})
+	}
+}
+
+func testAddAndListWithdrawals(t *testing.T, repo withdrawalRepository) {
+
```
altafan, line 37: Same here, and please remove blank line.

internal/infrastructure/storage/db/test/withdrawal_repository_impl_test.go
```diff
@@ -0,0 +1,108 @@
...
+		1,
+		domain.Page{
+			Number: 4,
+			Size:   10,
+		},
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+
```
altafan, line 81: we should also enforce that there is the expected number of withdrawals (100) in the repo.

internal/interfaces/grpc/handler/operator.go
```diff
@@ -194,6 +194,81 @@ func (o operatorHandler) ListWebhooks(
...
+func (o operatorHandler) listDeposits(
+	ctx context.Context,
+	req *pb.ListDepositsRequest,
+) (*pb.ListDepositsReply, error) {
+	deposits, err := o.operatorSvc.ListDeposits(
+		ctx,
+		int(req.GetAccountIndex()),
+		domain.Page{
+			Number: int(req.GetPage().GetPageNumber()),
+			Size:   int(req.GetPage().GetPageSize()),
```
altafan, line 220: This panics if the request does not contain any Page. Would be better to default to a fixed page in such case.

internal/core/domain/deposit_repository.go
```diff
@@ -0,0 +1,17 @@
+package domain
+
+import "context"
+
+type Page struct {
+	Number int
+	Size   int
+}
```
altafan, line 8: Would be better to be in a dedicated page_model.go file. Also, we should add a method Validate to this type that make sure that the page is valid: Number and Size MUST be both greater than 0 (badger repo would panic otherwise in such cases). This method should be ALWAYS called by any repo that uses Page to prevent unexpected errors/panics.

internal/infrastructure/storage/db/test/deposit_repository_impl_test.go
sekulicd: I think there is no need to check this since [line](https://github.com/sekulicd/tdex-daemon/blob/d508c727f0e9fa49148fb2ffefc203f068de5428/internal/infrastructure/storage/db/badger/deposit_repository_impl.go#L69) prevents insertion for same key(same in withdrawals repo)

internal/infrastructure/storage/db/test/withdrawal_repository_impl_test.go
sekulicd: Done

internal/infrastructure/storage/db/test/deposit_repository_impl_test.go
altafan: these tests are not only for badger but for any impl of the repo domain interfaces. I have no doubt badger impl is fine but let's check if also the inmemory impl sticks with this constraint.

sekulicd: agree

sekulicd: Done

internal/infrastructure/storage/db/test/withdrawal_repository_impl_test.go
sekulicd: Done

internal/interfaces/grpc/handler/operator.go
sekulicd: Actually this wont panic since protobuf pkg has validation, i added default values.
Done

internal/core/domain/deposit_repository.go
sekulicd: Done

tiero:
Missing commits re: your done comments? @sekulicd

sekulicd:
> Missing commits re: your done comments? @sekulicd

I done all comments. @altafan can u review?

tiero:
I mean, the latest commit I see d508c727f0e9fa49148fb2ffefc203f068de5428 seems older than comments. Maybe is github? or you need to force push

internal/interfaces/grpc/handler/operator.go
```diff
@@ -215,10 +215,10 @@ func (o operatorHandler) listDeposits(
 	deposits, err := o.operatorSvc.ListDeposits(
 		ctx,
 		int(req.GetAccountIndex()),
-		domain.Page{
-			Number: int(req.GetPage().GetPageNumber()),
-			Size:   int(req.GetPage().GetPageSize()),
-		},
+		domain.NewPage(
+			int(req.GetPage().GetPageNumber()),
```
altafan, line 219: This is still not panic safe! If in the request no page is defined, here GetPage() will return nil and trying to access GetPageNumber from a nil pointer will make the daemon to panic. In case GetPage() returns nil we simply create a default page domain like domain.NewPage(0,0).

sekulicd: @altafan now ok?

altafan: Actually no: this way if the page and number in the request are invalid they are still passed down because their validation is only done in domain.NewPage.
The code here should be like this IMO:
```
page := domain.NewPage(0,0)
if req.GetPage() != nil {
   page = domain.NewPage(
       int(req.GetPage().GetPageNumber()),
       int(req.GetPage().GetPageSize()),
   )
}
```

sekulicd: Done

altafan (approved):

PR #419, opened 2021-09-07
native asset

### LINKED ISSUE #408: Do not assume the Market Base asset is L-BTC
It can happen that a market's base asset happens to be something like a stablecoin or other asset that can't be used to pay for transaction fees. At the moment we assume that, ie. `operatorService` should have a parameter to know the asset hash of the blockchain used to pay for transaction fees

sekulicd:
This adds option to set lbtc asset hash of the network struct in regtest for the reasons explained in the issue itself.

It also removes unused marketBaseAsset property from blockchain_listener, wallet_service, walletunlocker_service.

This closes #408 

### DIFF
--- a/cmd/tdexd/main.go
+++ b/cmd/tdexd/main.go
@@ -32,7 +32,6 @@ import (
 var (
 	// General config
 	logLevel                = config.GetInt(config.LogLevelKey)
-	network                 = config.GetNetwork()
 	profilerEnabled         = config.GetBool(config.EnableProfilerKey)
 	datadir                 = config.GetDatadir()
 	dbDir                   = filepath.Join(datadir, config.DbLocation)
@@ -103,11 +102,20 @@ func main() {
 		log.Errorf("error while setting up webhook pubsub service: %s", err)
 		return
 	}
+
+	network, err := config.GetNetwork()
+	if err != nil {
+		crawlerSvc.Stop()
+		repoManager.Close()
+
+		log.Errorf("error while setting up network: %s", err)
+		return
+	}
+
 	blockchainListener := application.NewBlockchainListener(
 		crawlerSvc,
 		repoManager,
 		webhookPubSub,
-		marketsBaseAsset,
 		network,
 	)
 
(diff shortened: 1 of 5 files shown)


config/config.go
```diff
@@ -160,11 +164,22 @@ func GetBool(key string) bool {
 }
 
 //GetNetwork ...
-func GetNetwork() *network.Network {
+func GetNetwork() (*network.Network, error) {
```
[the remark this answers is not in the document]
sekulicd: Done

cmd/tdexd/main.go
```diff
@@ -103,11 +102,20 @@ func main() {
...
+
+	network, err := config.GetNetwork()
+	if err != nil {
+		crawlerSvc.Stop()
+		repoManager.Close()
+
+		log.Errorf("error while setting up network: %s", err)
+		return
+	}
+
```
altafan, line 114: instead of doing this here, we should check for the error in [validate()](https://github.com/tdex-network/tdex-daemon/blob/master/config/config.go#L202) func (right after the datadir check) which is called by init(). This way, here we could just skip checking for error.

sekulicd: this is cosmetic stuff and IMHO it is more readable if err is handled in this location and not skipped.

config/config.go
```diff
@@ -37,6 +39,8 @@ const (
 	NetworkKey = "NETWORK"
 	// BaseAssetKey is the default asset hash to be used as base asset for all markets. Default is LBTC
 	BaseAssetKey = "BASE_ASSET"
+	//NativeAssetHash is used to set lbtc hash, used for fee outputs, in regtest network
+	NativeAssetHash = "NATIVE_ASSET_HASH"
```
tiero, line 43: Just looking at it now,let be consistent

```suggestion
	NativeAssetHash = "NATIVE_ASSET"
```

tiero, line 43:
```suggestion
	NativeAssetKey = "NATIVE_ASSET_HASH"
```

tiero: need to use it everywhere in the code the new name `NativeAssetKey`

PR #422, opened 2021-09-09
Persist unspents

### LINKED ISSUE #411: Slice initialisation issue
Check slice init in [here](https://github.com/tdex-network/tdex-daemon/blob/70d50a8b72a45fbb78e2eb8bd428438090230f88/internal/core/application/operator_service.go#L1101)

This can lead to various bugs, since we are creating slice which length is not 0 but it is populated with empty structs.
Imagine ranging over this kind of slice and inserting in DB empty structs.

Maybe at this exact place we wont have bug but in general IMHO if we know (for sure) how many elements slice will have we should initialise like `make([]someSlice, 0, len(known_num))`

@tiero @altafan thoughts ?

### LINKED ISSUE #414: Move unspents to persistent filesystem to speed-up restarts
This is orthogonal to #413. We really need to understand how much we "loose" in terms of time to accept a TradePropose, moving the unspents (again) to Badger, rather than keep it in memory.

altafan:
With this, the daemon will persist in a physical file into the datadir the entries of the unspent repository. For backward compatibility, at start time, if a wallet is found in the db, then if there are no unspents the daemon attempts to restore them like it was doing before this PR. Alternatively, it just signals that it is already restored and skip the process above.

This also fixes ALL slice initializations that were using assignment instead of concatenation.

This closes #411.
This closes #414.

Please @tiero, review this.

### DIFF
--- a/internal/core/application/walletunlocker_service.go
+++ b/internal/core/application/walletunlocker_service.go
@@ -725,9 +725,9 @@ func (w *walletUnlockerService) restoreUnspentsForAddress(
 	}
 	utxos := iUtxos.([]explorer.Utxo)
 
-	unspents := make([]domain.Unspent, len(utxos), len(utxos))
-	for i, u := range utxos {
-		unspents[i] = domain.Unspent{
+	unspents := make([]domain.Unspent, 0, len(utxos))
+	for _, u := range utxos {
+		unspents = append(unspents, domain.Unspent{
 			TxID:            u.Hash(),
 			VOut:            u.Index(),
 			Value:           u.Value(),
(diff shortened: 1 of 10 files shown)


internal/core/application/walletunlocker_service.go
```diff
@@ -103,30 +103,37 @@ func newWalletUnlockerService(
...
-			); err != nil {
-				log.Infof("Failed for reason: %s", err)
-				w.setRestored(false)
-				return
+			// check whether the whole internal UTXO set needs to be restored
+			unspents := w.repoManager.UnspentRepository().GetAllUnspents(ctx)
+			if len(unspents) > 0 {
+				w.setRestored(true)
+			} else {
+				log.Info("Restoring internal wallet's utxo set. This could take a while...")
```
tiero, line 122:
```suggestion
				log.Info("Rescan internal wallet's utxo set. This could take a while...")
```

tiero (approved):

sekulicd (commented):
ACK

PR #433, opened 2021-09-14
Add missing CLI commands & Fixes

### LINKED ISSUE #425: Expose all missing operartor RPCs as `tdex` commands

### LINKED ISSUE #431: Start watching tx in SendToMany and WithdrawMarket RPCs
In these RPCs we MUST start watching the created tx to add/spend/confirm the unspent of the fee account and, in case of WithdrawMarket, also those of the market account.

We can reuse the code for watching transactions in the blockchain listener although it requires little modifications.

At the moment, using these 2 RPCs cause the utxo set to go out-of-sync.

### LINKED ISSUE #432: [Operator] Add balances info to ListMarket RPC reply

altafan:
This adds all the missing commands to the operator CLI:

* WalletUnlocker service
   * `status` calls the IsReady function and prints the reply.
   * `changepassword` to change the unlocking password. This required a small change to pkg/securestore in order to be able to change the pwd and list the internal buckets while locked.
* Operator service
   * `balancefee` returns the balance of the fee account.
   * `withdrawmarket` to withdraw funds from a market.
   * `reportmarketfee` to check the fees collected for a market.
* Wallet service (`wallet` cmd)
   * `receive` subcmd to retrieve a deposit address.
   * `balance` subcmd to check the balance of the Wallet account.
   *  `send` to send funds from the Wallet account to some address

This also contains some major and minor fixes:

* Add an `ExtraData interface{}` field to the tx observable and event and return a TransactionUnconfirmed only once in pkg/crawler
* Fix tx observation in application/blockchain_listener to not work with trades but with "common" transactions.
* Fix Operator and Wallet service to start watching for the created/broadcasted tx in `SendToMany` and `WithdrawMarket` to keep utxo set in sync.
* Fix the price returned in the `ReportMarketFee` RPC.
* Add balance info to reply of  `ListMarket`.
* Add tx id to reply of `SendToMany` and `WithdrawMarket` RPCs

This closes #425.
This closes #431.
This closes #432.

Please @tiero review this.


cmd/tdex/withdrawmarket.go
```diff
@@ -0,0 +1,81 @@
...
+	if err != nil {
+		return err
+	}
+	defer cleanup()
+
+	baseAsset, quoteAsset, err := getMarketFromState()
+	if err != nil {
+		return err
+	}
+	baseAmount := uint64(ctx.Float64("base_amount") * math.Pow10(8))
```
tiero, line 51: not always true we have 8 decimals. maybe is better to use satoshis, still a dev tool

internal/core/application/operator_service.go
```diff
@@ -650,6 +654,7 @@ func (o *operatorService) ListMarket(
 				FixedBaseFee:  market.FixedFee.BaseFee,
 				FixedQuoteFee: market.FixedFee.QuoteFee,
 			},
+			Balance: *balance,
```
tiero, line 657: shoudlnt getUnlockedBalanceForMarket return already the pointer to a balance struct?

altafan: this is to dereference the pointer actually since Balance is of type `Balance`, not of type `*Balance`.

PR #435, opened 2021-09-15
Add pagination for all listing RPCs & Update proto

### LINKED ISSUE #428: ListTrades: add pagination

### LINKED ISSUE #429: ReportMarketFee: add pagination

### LINKED ISSUE #430: ListUtxos: add pagination

altafan:
Before this, some listing RPCs were completely missing the possibility to pagine the response, and in those supporting this option, the only way to list all elements, like for example the deposits, was to define 1 page of a really big size to be sure to include all of them.

With this, instead, all listing methods support pagination and this becomes also optional, meaning that all RPCs supporting this feature, now return **ALL** the elements of the list in case the page is not defined in the incoming request.

This also contains some fixes:
* updates the `FeeInfo` message of the operator.proto by specifying both percentage fee amount and fixed fee amount as collected fees
* allows listing trades by market. Before it was only possible to list all trades related to all markets. The gRPC interface changes only by the fact that it is possible to pass a Market in the ListTradesRequest, while in the app service a dedicated method `ListTradesForMarket` has been added for the purpose.
* fixes updating trade in `TradeComplete`. Before this the changes to the trade domain were persisted only if no error occurred during the process. Now the updates are committed even in case of error because we must persist that the trade has failed.

This closes #428.
This closes #429.
This closes #430.

Please @tiero review this.


internal/core/domain/withdrawal_repository.go
```diff
@@ -4,10 +4,8 @@ import "context"
...
 	AddWithdrawal(ctx context.Context, withdrawal Withdrawal) error
-	ListWithdrawalsForAccountIdAndPage(
-		ctx context.Context,
-		accountIndex int,
-		page Page,
+	ListWithdrawalsForAccountId(
+		ctx context.Context, accountIndex int, page *Page,
 	) ([]Withdrawal, error)
-	ListAllWithdrawals(ctx context.Context) ([]Withdrawal, error)
+	ListAllWithdrawals(ctx context.Context, page *Page) ([]Withdrawal, error)
```
sekulicd, line 10: Based on name all objects should be retrieved and therefor Page object shouldnot be passed as arg.

internal/core/domain/unspent_repository.go
```diff
@@ -19,7 +19,7 @@ type UnspentRepository interface {
 	GetAvailableUnspents(ctx context.Context) ([]Unspent, error)
 	// GetAllUnspentsForAddresses returns the entire UTXO set (locked and spent
 	// included) for the provided list of addresses.
-	GetAllUnspentsForAddresses(ctx context.Context, addresses []string) ([]Unspent, error)
+	GetAllUnspentsForAddresses(ctx context.Context, addresses []string, page *Page) ([]Unspent, error)
```
sekulicd, line 22: Based on name all objects should be retrieved and therefor Page object shouldnot be passed as arg.

internal/core/domain/deposit_repository.go
```diff
@@ -4,10 +4,8 @@ import "context"
...
 	AddDeposit(ctx context.Context, deposit Deposit) error
-	ListDepositsForAccountIdAndPage(
-		ctx context.Context,
-		accountIndex int,
-		page Page,
+	ListDepositsForAccountId(
+		ctx context.Context, accountIndex int, page *Page,
 	) ([]Deposit, error)
-	ListAllDeposits(ctx context.Context) ([]Deposit, error)
+	ListAllDeposits(ctx context.Context, page *Page) ([]Deposit, error)
```
sekulicd, line 10: Based on name all objects should be retrieved and therefor Page object shouldnot be passed as arg.

internal/core/application/types.go
```diff
@@ -188,6 +189,19 @@ func (u Unspents) ToUtxos() []explorer.Utxo {
 	return l
 }
 
+type Page domain.Page
```
sekulicd, line 192: Looks like you wanted to remove domain structs from interface level, what was the motivation?

internal/core/domain/withdrawal_repository.go
altafan: you're right about the puprose of the method, but still, all withdrawals can be returned in a paginated way.

internal/core/application/types.go
altafan: the interface level should depend only by application in a direct way. Also, this "overridden" type is helpful to convert a page from interface level to the domain type in a clear way.

internal/core/domain/withdrawal_repository.go
altafan: same answer applies for comments below

tiero: I think we could have a little repetition but more clear

- ListAllWithdrawals
- ListPaginatedWithdrawals

altafan: Do you think it's more clear this way? consider also that this repetition would be needed for all other List/GetAll methods of the remaining repositories.

Another option could be to add comments (that will be part of the code documentation, therefore) describing the intended behavior for these methods.

tiero (approved):

PR #441, opened 2021-09-21
Add Withdraw fee account rpc & Refactor CLI commands & Add fields to customize wallet rescan range

### LINKED ISSUE #434: Persist deposit info also for market accounts
At the moment, the daemon persists only [the deposits relative to the fee account](../blob/master/internal/core/application/operator_service.go#L1238). Why this?

It should store also those related to markets, shouldn't it?

### LINKED ISSUE #439: tdex: expose tdex fee {balance, send, deposit} subcommands

### LINKED ISSUE #440: Add WithdrawFeeAccount RPC

### LINKED ISSUE #451: config: Add RESCAN_RANGE_START and RESCAN_RANGE_END

altafan:
This adds a new WithdrawFee RPC to withdraw funds from the fee account.

This also refactor the commands of the cli:
- the commands related to the fee account become now subcommand of the main `fee` command, similarly to the `wallet` one. So now, to deposit/claim/withdraw the funds the commands are `tdex fee deposit`, `tdex fee claim`, `tdex fee balance`  and the newly created `tdex fee withdraw`.
- the same refactor has been made for the original market commands. Now they're all grouped under the main `market` command and, like prior this, it's possible to manage the funds of the market (deposit/claim/withdraw), or manage its fees of (update fixed/percentage fee and report collected fees), strategy or price.

BONUS: This fixes the operator app service by storing also info about deposits of market accounts, not only those for the fee account.

NEW: This adds 2 new env vars RESCAN_RANGE_START and RESCAN_RANGE_END to customize the range of addresses to scan for each account when restoring the wallet in InitWallet.

This closes #434.
This closes #439.
This closes #440.
NEW: This closes #451.

Please @tiero review this.


tiero:
We should eventually maintain IMHO the old commands and flag as deprecated (dont recall if urfave has a way to do it, or just a Log.Warn that from.next release ie. Vo.7 will be removed)

tiero:
We can eventually just maintain the command that log.Info to use the new ome eventually. Maybe better since you deleted already and nor redo the work

altafan:
> We can eventually just maintain the command that log.Info to use the new ones. Maybe better since you deleted already the logic and not redo the work

@tiero Done.

cmd/tdex/claimfee.go
```diff
@@ -24,35 +19,8 @@ var claimfee = cli.Command{
...
-	return nil
-}
-
-func parseOutpoints(str string) ([]*pboperator.TxOutpoint, error) {
-	var outpoints []*pboperator.TxOutpoint
-	if err := json.Unmarshal([]byte(str), &outpoints); err != nil {
-		return nil, errors.New("unable to parse provided outpoints")
-	}
-	return outpoints, nil
+	return fmt.Errorf(
```
tiero, line 22: Use Warn maybe

internal/core/application/walletunlocker_service.go
```diff
@@ -504,11 +512,17 @@ func (w *walletUnlockerService) restoreAccount(
 	ww *wallet.Wallet, accountIndex int,
 ) *accountLastDerivedIndex {
 	lastDerivedIndex := &accountLastDerivedIndex{}
-	for chainIndex := 0; chainIndex <= 1; chainIndex++ {
+	ranges := [][2]int{
+		{w.rescanRangeStart, w.rescanRangeEnd}, // configurable range for external addresses
+		{0, 20},                                // fixed range for internal ones
```
tiero, line 517: shoudl come from env vars?

config/config.go
```diff
@@ -268,6 +280,18 @@ func validate() error {
...
+	end := GetInt(RescanRangeEndKey)
+	if start < 0 {
+		return fmt.Errorf("%s must not be a negative number", RescanRangeStartKey)
+	}
+	if end < 0 {
+		return fmt.Errorf("%s must not be a negative number", RescanRangeEndKey)
+	}
+	if start >= end {
+		return fmt.Errorf("%s must be greater than %s", RescanRangeStartKey, RescanRangeEndKey)
+	}
```
tiero, line 293: Wither we use env vars in config or RPC fields in InitWallet

internal/core/application/walletunlocker_service.go
altafan: this is the range for restoring internal addresses. This shouldn't be configurable because the user is not allowed to manage directly these addresses.

The range above is the one related to external address, which is indeed configurable via env var

PR #453, opened 2021-09-27
Refactor market RPCs

### LINKED ISSUE #450: Split deposit flow in many RPC
We should split the deposit flow in multiple RPCs, this has the benefit to not create empty markets by mistake and pave the way for future upgrades, like not forcing a single BASE_ASSET for all markets in a daemon.

```protobuf
rpc NewMarket(NewMarketRequest) returns (NewMarketReply) {}
rpc GetAddressForMarket() 
rpc GetAddressForFeeAccount()


message  NewMarketRequest {
  Market market = 1;
}
```

This should simplify the FundMarket in the domain code, since we assume the user to pass down the asset hashes he want to deposit

altafan:
This contains a refactor of the Operator RPCs that allow to manage a market.

The major change is a revisitation of the market creation flow:
Before this, one would have called the DepositMarket RPC without specifying a market in the request to create a new market and then the asset pair would be inferred when claiming the funds.
With this, instead, the operator MUST create a market with the new NewMarket RPC. It is mandatory to specify the asset pair of the market in the request. When claiming the funds, the utxo set for the market account is updated by making sure that all funds are of one of the asset pair hashes.

The minor changes are about the renaming of some existing RPCs to make them more meaningful:
- DepositFeeAccount -> GetFeeAddress
- ListFeeExternalAddresses (new)
- FeeAccountBalance -> GetFeeBalance
- ClaimFeeDeposit -> ClaimFeeDeposits
- NewMarket (new)
- DepositMarket -> GetMarketAddress
- GetMarketBalance (new)
- ClaimMarketDeposit -> ClaimMarketDeposits
- ListMarket -> ListMarkets

This closes #450.

Please @tiero review this.


internal/core/application/walletunlocker_service.go
```diff
@@ -609,7 +602,28 @@ func (w *walletUnlockerService) restoreMarket(
...
-		if err := market.FundMarket(outpoints, w.marketBaseAsset); err != nil {
+		outpointsByAsset := make(map[string]struct{})
+		for _, o := range outpoints {
+			outpointsByAsset[o.Asset] = struct{}{}
+		}
+		var baseAsset, quoteAsset string
+		for asset := range outpointsByAsset {
+			if asset == w.marketBaseAsset {
+				baseAsset = asset
+			} else {
```
tiero, line 613: what will happens if my mistake the user deposits 3 assets types? we should eventually check explicitly for quoteAsset here

internal/infrastructure/storage/db/test/market_repository_impl_test.go
```diff
@@ -352,15 +328,15 @@ func testWriteRollback(t *testing.T, repo marketRepository) {
...
 	return []marketRepository{
-		{
-			Name:       "badger",
-			DBManager:  badgerDBManager,
-			Repository: badgerDBManager.MarketRepository(),
-		},
+		// {
+		// 	Name:       "badger",
+		// 	DBManager:  badgerDBManager,
+		// 	Repository: badgerDBManager.MarketRepository(),
```
tiero, line 338: why commenting this out?

api-spec/protobuf/operator.proto
```diff
@@ -7,87 +7,90 @@ import "types.proto";
...
+  // Makes the given market NOT tradabale.
   rpc CloseMarket(CloseMarketRequest) returns (CloseMarketReply) {}
 
-  // Get extended details for each markets either open, closed or to be funded.
-  rpc ListMarket(ListMarketRequest) returns (ListMarketReply) {}
+  // Deletes a market.
+  rpc DropMarket(DropMarketRequest) returns(DropMarketReply){}
+
+  // Displays a report of the colected fees for the given market.
+  rpc ReportMarketFee(ReportMarketFeeRequest) returns (ReportMarketFeeReply) {}
```
tiero, line 59: I would modify also this RPC to make it more clear which `fee` we are talking about, to not clash with `UpdateMarket{Percentage|Fixed}Fee` meaning 

```suggestion
  rpc ReportSwapFeesForMarket(ReportMarketFeeRequest) returns (ReportMarketFeeReply) {}
```

internal/core/application/walletunlocker_service.go
altafan: Here we're in the process of restoring a daemon wallet and because depositing 3 asset types in a market is an operation not allowed by the daemon, we suppose here that all the markets were previously created in the right way.

If we don't want to be so optimistic, here the only thing we can do is returning an error because the daemon cannot be sure of the asset to use the quote one of the market.

internal/infrastructure/storage/db/test/market_repository_impl_test.go
altafan: ops, left-over 🙏

internal/core/application/walletunlocker_service.go
altafan: > we should eventually check explicitly for quoteAsset here

This is not possible at this stage. In this function, we're restoring the markets, which means that we don't know the quote assets of each market found and we need to discover them. 
`w.marketBaseAsset` is basically the configurable base asset from env var. It's really helpful in this process because it allows the daemon to figure out the market pairs of every market account to be restored.

cmd/tdex/fee.go
```diff
@@ -36,6 +36,11 @@ var (
 		},
 		Action: feeDepositAction,
 	}
+	feeListAddressesCmd = &cli.Command{
+		Name:   "listaddresses",
```
tiero, line 40:
```suggestion
		Name:   "addresses",
```

cmd/tdex/market.go
```diff
@@ -37,13 +37,30 @@ var (
...
+		Flags: []cli.Flag{
 			&cli.IntFlag{
 				Name:  "num_of_addresses",
 				Usage: "the number of addresses to generate for the market",
 			},
 		},
 		Action: marketDepositAction,
 	}
+	marketListAddressesCmd = &cli.Command{
+		Name:   "listaddresses",
```
tiero, line 60:
```suggestion
		Name:   "addresses",
```

PR #454, opened 2021-09-30
Bader DB: Use Find() and UpdateMatching() 

### LINKED ISSUE #452: Unspents not locked after a trade
I found an interesting bug that seems to happen only after restarting a daemon:

Right after the daemon accepted a trade, the unspent used as inputs of the tx should be marked as locked.

While this seems to work properly for a running daemon that has never been stopped, I found out from the debug messages that by restarting a daemon, unlocking, and making a trade against it causes the daemon to not lock any of the unspents used in the accepted trade.

The following are logs of test daemon in regtest network:
```
DEBU[0038] /Trade/TradePropose
INFO[0038] trade with id 45809343-7301-4bc7-9901-bc717e216d14 accepted
DEBU[0038] locked 0 unspents
DEBU[0038] /Trade/TradeComplete
INFO[0038] trade with id 45809343-7301-4bc7-9901-bc717e216d14 completed
WARN[0038] unable to broadcast trade tx                  error="sendrawtransaction RPC error: {\"code\":-25,\"message\":\"Missing inputs\"}" hex=....
```

The result of this misbehavior is the construction of an invalid transaction that is rejected by the network when broadcasted.

altafan:
This fixes the badger implementation of the unspent repository so that it doesn't make use of badgerhold's methods like `Get()` or `Update()` to read/write data from/to the store. They are replaced by `Find()` and `UpdateMatching()` that actually query the store instead of getting/updating an elem by its key.

There is still an [open issue](https://github.com/timshannon/badgerhold/issues/76) on the badgerhold repo for this. These changes could eventually be rolled back if we figure out the reason of such weird misbehavior.

This closes #452.

BONUS:
- This fixes the trade domain to prevent being able to settle a trade if it's failed (a dedicated test case has been added to prove it)
- This fixes the bc listener to properly stop the outpoint observation if the transaction has failed for any reason.

Please @tiero review this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,12 +20,8 @@ jobs:
       - name: Check out code into the Go module directory
         uses: actions/checkout@v2
 
-      - name: Install Nigiri
-        run: |
-          mkdir ~/.nigiri; cd ~/.nigiri
-          curl https://travis.nigiri.network | bash; cd
-          docker-compose -f ~/.nigiri/docker-compose.yml up -d
-          sleep 5
+      - name: Run Nigiri
+        uses: vulpemventures/nigiri-github-action@v1
 
       - name: Get dependencies
         run: go get -v -t -d ./...
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -18,12 +18,8 @@ jobs:
       - name: Check out code into the Go module directory
         uses: actions/checkout@v2
 
-      - name: Install Nigiri
-        run: |
-          mkdir ~/.nigiri; cd ~/.nigiri
-          curl https://travis.nigiri.network | bash; cd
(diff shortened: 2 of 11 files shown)


tiero:
> This fixes the bc listener to properly stop the outpoint observation if the transaction has failed for any reason.

What if the node/explorer is not reachable for that moment, but the user will the retry later? Maybe better having a timeout?

tiero:
Failing test `TestFailingTradeSettle`

altafan:
> What if the node/explorer is not reachable for that moment, but the user will the retry later? Maybe better having a timeout?

The observation of the outpoints starts BEFORE the broadcasting of the transaction, which, better to remember, can be done by the daemon via TradeComlete or by the trader on his own. Not sure I got the timeout thing.

Issue #456, opened 2021-09-30
Pluggable: BasePrice and QuotePrice can be less than zero


tiero:
In this commit we introduced a validation check on the Price domain https://github.com/tdex-network/tdex-daemon/commit/ebfe465a1c8b5f111385a7383d55f475e94f6a35#diff-4d41ed4b84fbbf1a8d096677eb4767a6976b057cfa37f81e89111b502a488390R126-R131 that is conceptually wrong: the prices can def be less than zero

tiero:
We should check instead they are at least one satoshi tho

altafan:
Closed by #457.

PR #457, opened 2021-10-01
HotFix: price validation in gRPC handler

altafan:
This fixes the validation of the price in the request of UpdateMarketPrice at interface level.

Please @tiero review this.

### DIFF
--- a/internal/interfaces/grpc/handler/operator.go
+++ b/internal/interfaces/grpc/handler/operator.go
@@ -987,8 +987,11 @@ func parsePrice(p *pbtypes.Price) (price application.Price, err error) {
 		basePrice = decimal.NewFromFloat32(p.GetBasePrice())
 		quotePrice = decimal.NewFromFloat32(p.GetQuotePrice())
 	}
-	pp := application.Price{basePrice, quotePrice}
-	if err = price.Validate(); err != nil {
+	pp := application.Price{
+		BasePrice:  basePrice,
+		QuotePrice: quotePrice,
+	}
+	if err = pp.Validate(); err != nil {
 		return
 	}
 	price = pp


internal/interfaces/grpc/handler/operator.go
```diff
@@ -987,8 +987,11 @@ func parsePrice(p *pbtypes.Price) (price application.Price, err error) {
 		basePrice = decimal.NewFromFloat32(p.GetBasePrice())
 		quotePrice = decimal.NewFromFloat32(p.GetQuotePrice())
 	}
-	pp := application.Price{basePrice, quotePrice}
-	if err = price.Validate(); err != nil {
+	pp := application.Price{
+		BasePrice:  basePrice,
+		QuotePrice: quotePrice,
+	}
+	if err = pp.Validate(); err != nil {
```
tiero, line 994: 🤔

tiero: if we pass a price like 0.001 will still return (wrongly) the error

This check has been introduced in latest rc, is the issue
https://github.com/tdex-network/tdex-daemon/blob/master/internal/core/application/types.go#L127

tiero: https://github.com/tdex-network/tdex-daemon/issues/456

altafan: no 0.001 is not lower than 0 and this is the right fix: the problem was that we were validating `price`, which is the return value of the `parsePrice` function, instead of validating `pp` which is the price of the incoming request.

tiero: 🙏  you right

PR #458, opened 2021-10-01
Add rate limiter to esplora GetUnspentsForAddresses & Fix rescan of external/internal addresses when restoring

### LINKED ISSUE #455: Explorer: Rate limit & postpone
The pkg explorer (which is consumed by other Go projects) should have a rate limiter with a maximum number of calls to perform per second, with a bucket (queue) to keep the "limited" one waiting to be executed.

This packages seems handy https://github.com/uber-go/ratelimit or just a tweak to the `rate` pkg code we use in the crawler.

altafan:
This removes the usage of a rate limiter from the Cralwer service since it is not actually needed.

At the same time, this adds the usage of a leaky bucket rate limiter to the esplora implementation of the  GetUnspentsForAddresses API in order to prevent flooding the explorer with too many requests in very short time.

Since now the GetUnpsentsForAddresses can be used to retrieve the utxos for a long list of addresses, it is consumed when rescanning the utxo set in the InitWallet+restore flow.

Also, the `RESCAN_RANGE_END` env var has been renamed `RESCAN_GAP_LIMIT.`
The gap limit is applied not only for external addresses, but also for internal ones instead of hardcoding it to `20`.
The `RESCAN_RANGE_START`, instead, is applied only to external addresses.

This fixes #455.

Please @tiero review this.


cmd/tdexd/main.go
```diff
@@ -51,13 +51,11 @@ var (
 	tradeSvcPort                  = config.GetInt(config.TradeListeningPortKey)
 	operatorSvcPort               = config.GetInt(config.OperatorListeningPortKey)
 	crawlerIntervalInMilliseconds = time.Duration(config.GetInt(config.CrawlIntervalKey)) * time.Millisecond
-	crawlerLimit                  = config.GetInt(config.CrawlLimitKey)
-	crawlerTokenBurst             = config.GetInt(config.CrawlTokenBurstKey)
 	explorerTimoutRequest         = config.GetDuration(config.ExplorerRequestTimeoutKey)
 	cbMaxFailingRequest           = config.GetInt(config.CBMaxFailingRequestsKey)
 	cbFailingRatio                = config.GetFloat(config.CBFailingRatioKey)
 	rescanRangeStart              = config.GetInt(config.RescanRangeStartKey)
```
tiero, line 57: Semantically is not anymore a range, so if we drop end, we should just call it `RESCAN_START`

config/config.go
```diff
@@ -127,13 +121,11 @@ func init() {
...
 	vip.SetDefault(EnableProfilerKey, false)
 	vip.SetDefault(StatsIntervalKey, 600)
-	vip.SetDefault(CrawlLimitKey, 10)
-	vip.SetDefault(CrawlTokenBurstKey, 1)
 	vip.SetDefault(NoMacaroonsKey, false)
 	vip.SetDefault(CBMaxFailingRequestsKey, 20)
 	vip.SetDefault(CBFailingRatioKey, 0.7)
 	vip.SetDefault(RescanRangeStartKey, 0)
-	vip.SetDefault(RescanRangeEndKey, 20)
+	vip.SetDefault(RescanGapLimitKey, 20)
```
tiero, line 128: Let's put something like 50

pkg/explorer/esplora/unspents.go
```diff
@@ -36,10 +36,11 @@ func (e *esplora) GetUnspentsForAddresses(
 		close(chRes)
 	}()
 
+	limiter := ratelimit.New(len(addresses))
 	for i := range addresses {
 		addr := addresses[i]
-		go e.getUnspentsForAddress(addr, blindingKeys, chRes, wg)
-		time.Sleep(1 * time.Millisecond)
+		go e.getUnspentsForAddress(addr, blindingKeys, chRes, wg, limiter)
+		// time.Sleep(1 * time.Millisecond)
```
tiero, line 43: remove if not needed anymore

PR #471, opened 2021-10-07
Add Tdexdconnect binary

### LINKED ISSUE #466: tdexconnect
Following the example of [`lndconnect`](https://github.com/LN-Zap/lndconnect), we should add a new binary `tdexconnect`, that would be useful for the apps that want to interact with the Operator interface without sourcing eventually macaroons and TLS certificates by hand.

altafan:
This adds a new deliverable`tdexdconnect` to the repository. This program encodes daemon's connection params (RPC address, chai,  TLS certificate and macaroon) into a URL string that can be decoded by other apps to facilitate their configuration in order to interact with the daemon.

_tdexdconnect_ accepts 4 flags to customize the URL encoding:
* `--rpcserver` to set the daemon's addr to connect to (defaults to localhost:9000)
* `--tls_cert_path` to change the default path where to find the certificate (daemon's default datadir)
* `--macaroons_path` to change the default path where to find the macaroon (daemon's default datadir). In case this flag is set, it's mandatory for the file to exist. If, instead, the default is used, it is not added to the URL if the file doesn't exist.
* `--network` the chain (regtest, liquid) set for the daemon
* `--out` the type of output produced, either a string, a QRCode on stdout or a QRCode in a .png file

Since the operator CLI is one of the apps mentioned before, a new subcommand `config connect` has been added to make use of a tdexdconnect URL to configure it instead of using `config init` for example.

This closes #466.

Please @tiero review this.

### DIFF
--- a/cmd/tdex/config.go
+++ b/cmd/tdex/config.go
@@ -196,7 +196,7 @@ func configConnectAction(c *cli.Context) (err error) {
 		}
 	}()
 
-	noMacaroons := strconv.FormatBool(certificate == nil && macaroon == nil)
+	noMacaroons := strconv.FormatBool(len(certificate) == 0 && len(macaroon) == 0)
 
 	if err = setState(map[string]string{
 		"rpcserver":      rpcServerAddr,
--- a/cmd/tdex/initwallet.go
+++ b/cmd/tdex/initwallet.go
@@ -5,6 +5,8 @@ import (
 	"encoding/hex"
 	"fmt"
 	"io"
+	"io/ioutil"
+	"path/filepath"
 	"strings"
 
 	pbwallet "github.com/tdex-network/tdex-daemon/api-spec/protobuf/gen/walletunlocker"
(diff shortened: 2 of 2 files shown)


pkg/tdexdconnect/tdexdconnect.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	certificate := base64.RawURLEncoding.EncodeToString(block.Bytes)
+	q.Add("cert", certificate)
+
+	if macBytes != nil {
+		macaroon := base64.RawURLEncoding.EncodeToString(macBytes)
+		q.Add("macaroon", macaroon)
+	}
+
+	q.Add("net", network)
```
tiero, line 29: what's this about

tiero: If we move the fragmenter features inside the daemon, the client (CLI or UI) do not need to handle any network check?

tiero (approved):

PR #477, opened 2021-10-22
Base trade expiration on blockheight & Spend utxos of expired trades

### LINKED ISSUE #361: "Double spend" expired trades
## Problem

When the daemon accepts a trade, is up to the trader to **complete** (ie. counter-sign) and either a) call the TradeComplete b) broadcast by himself to the blockchain directly.

This is an indeed a problem regarding the price of a trading pair: the trader can "hold" on purpose the trade, to broadcast let's say an hour later when the price is more favorable. To mitigate this, we added a `expiry_time_unix` to the `TradeProposeReply` to signal the trader that after that timestamp, we will DOUBLE SPEND the promised utxos he has in the accepted trade. At the moment WE ARE NOT DOING ANYTHING about expiration.

## Solutions

We should have in places a system that will programmatically try to double spend utxos being included in swap accepted messages and delivered to proposers.

Double spending costs L-BTCs, so is not a "cheap" operation, so we should try to use the "expired utxos" for upcoming trades if possible: the coin selection will always pick first utxos of previous expired transactions we have in a "garbage bin" queue, sort of "recycling" operation, but this cannot be relayed upon fully (especially if there aren't much trades happening), therefore we should have  a system that programmatically double spend it to the daemon itself.

altafan:
This introduces some changes to the logic of trade expiration:

1. With this, the expiration is not only based on current timestamp, but also on the height of the blockchain. Basically, to make a trade expiring, it's not enough that *now* is after some trade expiration date, but also that at least 1 new block has been mined in the blockchain.
For this, a new field `ProposalTime` has been added to the Trade domain, set at proposal time. Requires the caller to fetch the current block height from the blockchain.
2. When making a trade expiring, the utxos used as inputs are transferred to some new external address (maybe better to use internal?). In case the operation is not allowed because it's "too early", meaning that the expiration date  has not passed or no new blocks have been fetch in the blockchain, the operation is postponed of 2 minutes (the time window for a new block in  Liquid main chain).

Closes #361.

Please @tiero review this.


altafan:
@tiero what about this?

tiero:
I think we need much more talking about this: this expiration consumes fees and breaks the unspents even more.

In a real high frequency scenario, would be the coin selection to pick for a queue of "to be invalidated trades" first? ofc timing is important and we need to understand a nice way to add this, which is def more moving pieces.

PR #485, opened 2021-11-04
Add event info to payloads of webhook service

### LINKED ISSUE #481: Webhook: Include the event type inside the event payload

altafan:
This closes #481.

Please @tiero review this.


tiero (commented):
what about other topics, account low balance and withdraw @altafan

altafan:
i've added the event to the payloads of all types of topic @tiero.

tiero:
ah ok wait, before trade settled was in blockshain listner, was missing the function

altafan:
yeah, i added the util function just like did for other topics.. better to read like this.

PR #502, opened 2021-11-12
Use vault accounts for fragmenter

### LINKED ISSUE #495: Use vault reserved accounts instead of ephemeral wallet for fee/market fragmenter
At the moment, the fragmenters of both fee and markets accounts make use of ephemeral single-key wallets stored in-memory that are discarded after a fragmentation either succeeds or is aborted.

The problem with this solution are:
* an user could send funds to the address of a no-longer used ephemeral wallet by mistake and those funds would be lost forever.
* if the user restarts the daemon any eventual ephemeral wallet that was used before the restart is  lost foverer, and again, the user could loose funds forever.

Instead of using ephemeral single-key wallets for the fragmenters, we could use 2 of the reserved accounts of the vault domain. This way we could get rid of the problems described above.
We would use one account as the FeeFragmenterAccount, and another for the MarketFragmenterAccount.

altafan:
This updates the fragmentation logic in order to deprecate the usage of ephemeral wallets in favour of reserved accounts of Vault domain.

## Daemon

For this, the fragmenter portable interface (ports.Fragmenter) and all other relative types/interfaces have been dropped. Basically, the only thing re-used is the fragmentation logic of the original `FragmentAmount()` method of both fee and market fragmnenter implementations.
These methods have been moved to a `fragmenter_utils.go` file and became internal methods of the app layer.

As already introduced, this makes use of 2 Vault reserved accounts that have been named _FeeFragmenter_ and _MarketFragmenter_ accounts. This has the benefit of not taking the risk of loosing funds during a fragmentation as explained in the related ticket.

Several RPCs have been added for these 2 new accounts to look similar to those defined for the Fee and Market ones:
- `ListFeeFragmenterAddresses` to list all external addresses derived for the account
- `GetFeeFragmenterBalance` to retrieve the current balance of the account
- `FeeFragmenterSplitFunds` to perform fragmentation and deposit funds into fee account. The original `FragmentFeeFragmenterDeposits` is dropped.
- `WithdrawFeeFragmenter` to withdraw **ALL** the funds owned by the account to some address

The same is done for the MarketFragmenter account.

## CLI

On the CLI side, the interactive process originally done with `tdex fee|market deposit --fragment` is deprecated.
Instead, 2 new main commands are added:
- `feefragmenter` to manage the fee fragmenter account
- `marketfragmenter` to manage the market fragmenter account

Both commands have the following subcommands:
- `deposit [--num_of_addresses]`  to retrieve address where to send funds
- `addresses` to list all deposit (external) addresses
- `balance` to see the current balance of the account
- `split` to perform fragmentation&deposit into fee/market account 
- `withdraw` to send **ALL** the funds owned by the fragmenter to an external address

Therefore, the process is no-longer interactive. The operator needs to call at least `deposit` and `split` to perform the fragmentation of the funds.

### BONUS:
* In order to use the _pkg/wallet_ to craft the deposit or withdraw transaction of the fragmenters, it's been necessary to add the option to the `UpdateTx()` to subtract fees from some LBTC output that is going to be added to the transaction. This can now be done by setting the new param `SubtractFees`, which is mutually exclusive to `WantChangeForFees`, which instead instructs to either subtract the fees from a change LBTC out, or to add one by making a new coinselction.
* Rename the internal methods `sendToMany` and `sendToManyFeeAccount` of the app layer to `sendToManyWIthFeeTopup` and `sendToManyWithoutFeeTopup` to better understand their purpose.
* This fixes some typos in the operator.proto file

Closes #495.

Please @tiero review this.


tiero:
Just asking, Why we couldn't keep the `tdex fee|market deposit --fragment` UX instead of adding specific commands?

altafan:
Since FeeFragmenter and MarketFragmenter became accounts similar to the Fee and Market ones, I thought they needed similar RPCs for consistency. Like for instance to check the balance or to list of all derivated deposit addresses.

At this point, the CLI would have been enriched with new commands anyways, so I decided to deprecate the interactive process of the fee/market account commands.

tiero (approved):

Issue #501, opened 2021-11-12
[UpdateMarketFixedFee] Error: fixed fee requires both base and quote amounts to be defined


Janaka-Steph:
If setting only base or quote to value other than 0 an error `fixed fee requires both base and quote amounts to be defined` is returned. 
<img width="1025" alt="Screenshot 2021-11-12 at 13 14 27" src="https://user-images.githubusercontent.com/4022128/141473004-a2d989cf-ec50-4046-baa1-b8466e90d448.png">

tiero:
ah-ha, I need we check for 0 to see if is nil/undefined on in interface/protobuf level, we should accpet 0/nil as valid isn't? @altafan

altafan:
definitely

PR #503, opened 2021-11-15
Fix pluggable strategy type & Allow only base/quote fixed fee when updating

### LINKED ISSUE #500: When selecting Pluggable strategy, MarketInfo strategyType is 255
When selecting Pluggable strategy, MarketInfo strategyType is 255, instead of 0.

altafan:
This introduces changes to fix the strategy type returned by the daemon and to not require both the base and the quote fee to be defined to update the market fixed fee.

Closes #500.
Cloeses #501.

Please @tiero review this.

### DIFF
--- a/internal/core/domain/market_service.go
+++ b/internal/core/domain/market_service.go
@@ -20,18 +20,6 @@ func (p Prices) AreZero() bool {
 	return decimal.Decimal(p.BasePrice).Equal(decimal.NewFromInt(0)) && decimal.Decimal(p.QuotePrice).Equal(decimal.NewFromInt(0))
 }
 
-func (f FixedFee) validate() error {
-	if f.BaseFee < 0 || f.QuoteFee < 0 {
-		return ErrInvalidFixedFee
-	}
-
-	if (f.BaseFee > 0 && f.QuoteFee == 0) || (f.QuoteFee > 0 && f.BaseFee == 0) {
-		return ErrMissingFixedFee
-	}
-
-	return nil
-}
-
 // IsTradable returns true if the market is available for trading
 func (m *Market) IsTradable() bool {
 	return m.Tradable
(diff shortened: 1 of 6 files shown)


internal/core/domain/market_service.go
```diff
@@ -52,10 +40,8 @@ func (m *Market) QuoteAssetPrice() decimal.Decimal {
 }
 
 // IsStrategyPluggable returns true if the the startegy isn't automated.
-// For backward compatibility it returns whether the strategy is zero-ed or,
-// on the contrary, its type is extaclty PluggableStrategyType.
 func (m *Market) IsStrategyPluggable() bool {
-	return m.Strategy.IsZero() || m.Strategy.Type == PluggableStrategyType
+	return !m.Strategy.IsZero() && m.Strategy.Type == int(StrategyTypePluggable)
```
tiero, line 44: what is the impact of this change?

altafan: Since the type defined for the Pluggable strategy in the proto  is `0`, we need to be sure that in case the market strategy type has this kind of value, its formula must not be nil because it would mean that no strategy has been set for the current market. This is the impact of the change.

Before this, since `IsZero()` was just checking the type and not also the formula, the workaround was to use the `255` value for the pluggable strategy type (which generated #500).

The other change that contributes to fixing that ticket is [this](https://github.com/altafan/tdex-daemon/blob/fix-fixed-fee-and-strategy/internal/infrastructure/storage/db/badger/market_repository_impl.go#L366): before this, in the badger repo we were restoring only the non-pluggable strategies, this was due to the fact that in the early versions, Pluggable was not an actual `MakingFormula` implementation but rather, all its logic was hardcoded in the application level.
Now, instead, we restore all kinds of strategies, and evenmore, using the `default` branch for the pluggable one makes it compatible with markets storing either the `255` or the `0` type.
The very first change to the market, like for example a change of the price by some feeder, will then make the pluggable strategy type to be set to the correct `0` value for that market.

PR #515, opened 2021-11-29
Fragmenters: Wait for tx to be returned by the explorer before claiming & Use fee account to cover fees of market fragmentation

### LINKED ISSUE #514: [FragmenterSplitFunds] Claiming process should poll to wait for mempool transaction propagation
Working on Tdex Dashboard Testnet support.

I run the daemon with this command:
```
docker run -it -d --name tdexd \
-p 9945:9945 -p 9000:9000 \
-v `pwd`/tdexd:/home/user/.tdex-daemon \
-e TDEX_NO_MACAROONS=false \
-e TDEX_NETWORK="testnet" \
-e TDEX_BASE_ASSET="144c654344aa716d6f3abcc1ca90e5641e4e2a7f633bc09fe3baf64585819a49" \
-e TDEX_LOG_LEVEL=5 \
-e TDEX_FEE_ACCOUNT_BALANCE_THRESHOLD=1000 \
-e TDEX_EXPLORER_ENDPOINT="https://blockstream.info/liquidtestnet/api" \
ghcr.io/tdex-network/tdexd:v0.8.0-rc.1
```

I attempt to send some Testnet funds to the Fee account.

Sending 0.001 tL-BTC from my Green wallet to fee deposit address:
https://blockstream.info/liquidtestnet/tx/bc86c3ade6e7d041ae44d4467bc924f4717ee65ae22c58c199be9e835eeab4b8#blinded=1000000,144c654344aa716d6f3abcc1ca90e5641e4e2a7f633bc09fe3baf64585819a49,5e282fa594ece024012c5c59c254953ebad8d4b351362aade259e555b68da83c,45e22c2a14875cf5f5f6b887de03bcfec2baf024c16083dc11243e0801b21000,899743,144c654344aa716d6f3abcc1ca90e5641e4e2a7f633bc09fe3baf64585819a49,c34ffe4307bf327b0b39b5f2971940c4659193ef36e7d3f294eefd31befacc17,49e835336fbb6107955b9815337a5f9fb159105d6aacc00ceff7b4bec3203f98  

Fragmentation Tx:
https://blockstream.info/liquidtestnet/tx/bd1b708695b8966083d18706b9ed92e88f39fa34441ead7efeff7000b6803c46

Even though it has been mined, Tdex daemon returned an error `failed to broadcast transaction: sendrawtransaction RPC error: {\"code\":-25,\"message\":\"bad-txns-inputs-missingorspent\"}`.

On second fragmentation try I got `no funds detected for fee fragmenter`.

```
$ tdex feefragmenter balance
{
        "balance": {
        }
}
```

Funds are in the limbos.

### LINKED ISSUE #517: [Fragmenter] Market fragmentation must use the fee account to pay for split/withdraw fees
Since it can happen to have a market with two non-LBTC assets (USDT-LCAD), we need to not assume the base asset is LBTC and use then the fee account to pay for fees.

PS: We should be more explicit in the application error we throw in case an user is trying to split/withdraw from the fragmenter account(s) and the fee account balance is too low.

altafan:
This adds a waiting function that polls the explorer until the fee/market funding transaction made with the related fragmenter is returned.

Only after this, the fragmenter can proceed with claiming the funds for the account.

Closes #514.
Closes #517.

~~EDIT: This adds also a check to prevent splitting too small amounts in fee fragmentation~~

Please @tiero review this.

### DIFF
--- a/internal/core/application/explorer_utils.go
+++ b/internal/core/application/explorer_utils.go
@@ -2,6 +2,8 @@ package application
 
 import (
 	"context"
+	"strings"
+	"time"
 
 	"github.com/tdex-network/tdex-daemon/internal/core/ports"
 	"github.com/tdex-network/tdex-daemon/pkg/circuitbreaker"
@@ -62,4 +64,19 @@ func getBalancesByAsset(unspents []explorer.Utxo) map[string]BalanceInfo {
 		balances[unspent.Asset()] = balance
 	}
 	return balances
+}
+
+func waitForTx(explorerSvc explorer.Service, txid string) error {
+	waitingTime := 1 * time.Second
+	for {
+		_, err := explorerSvc.GetTransaction(txid)
+		if err != nil {
+			if strings.Contains(err.Error(), "Transaction not found") {
+				time.Sleep(waitingTime)
+				continue
+			}
+			return err
+		}
+		return nil
+	}
 }
--- a/internal/core/application/explorer_utils.go
+++ b/internal/core/application/explorer_utils.go
@@ -2,8 +2,6 @@ package application
 
 import (
 	"context"
-	"strings"
-	"time"
 
 	"github.com/tdex-network/tdex-daemon/internal/core/ports"
 	"github.com/tdex-network/tdex-daemon/pkg/circuitbreaker"
@@ -64,19 +62,4 @@ func getBalancesByAsset(unspents []explorer.Utxo) map[string]BalanceInfo {
 		balances[unspent.Asset()] = balance
 	}
 	return balances
-}
-
-func waitForTx(explorerSvc explorer.Service, txid string) error {
-	waitingTime := 1 * time.Second
-	for {
-		_, err := explorerSvc.GetTransaction(txid)
-		if err != nil {
-			if strings.Contains(err.Error(), "Transaction not found") {
-				time.Sleep(waitingTime)
-				continue
-			}
-			return err
-		}
-		return nil
-	}
 }
--- a/internal/core/application/operator_service.go
+++ b/internal/core/application/operator_service.go
@@ -35,7 +35,8 @@ var (
 		3: 10,
 		5: 2,
 	}
-	PollInterval = 1 * time.Second
+	PollInterval           = 1 * time.Second
+	MinFeeFragmenterAmount = 5000
 )
 
 // OperatorService defines the methods of the application layer for the operator service.
@@ -1839,6 +1840,15 @@ func (o *operatorService) splitFeeFragmenterFunds(
 			Err: fmt.Errorf(
 				"found funds with asset different from LBTC. Use a recover address " +
 					"to get them back",
+			),
+		}
+		return
+	}
+	if int(amountPerAsset[lbtc]) < MinFeeFragmenterAmount {
+		chRes <- FragmenterSplitFundsReply{
+			Err: fmt.Errorf(
+				"amount to fragment is too small, should be at least %d sats of LBTC",
+				MinFeeFragmenterAmount,
 			),
 		}
 		return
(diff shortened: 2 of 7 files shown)


internal/core/application/explorer_utils.go
```diff
@@ -63,3 +65,18 @@ func getBalancesByAsset(unspents []explorer.Utxo) map[string]BalanceInfo {
 	}
 	return balances
 }
+
+func waitForTx(explorerSvc explorer.Service, txid string) error {
+	waitingTime := 1 * time.Second
+	for {
+		_, err := explorerSvc.GetTransaction(txid)
+		if err != nil {
+			if strings.Contains(err.Error(), "Transaction not found") {
```
tiero, line 74: This could be fragile. Better to check the HTTP status code ie. 404? (if is that)

tiero: ie. if the text of the error changes

altafan: Wait, the solution is wrong, ie explorer is a general interface while we check against a specific error of the esplora impl.
I'm going to add a new method to the explorer `PollGetKnowTransaction` to the explorer.

altafan: @tiero done.

internal/core/application/operator_service.go
```diff
@@ -1843,6 +1844,15 @@ func (o *operatorService) splitFeeFragmenterFunds(
 		}
 		return
 	}
+	if int(amountPerAsset[lbtc]) < MinFeeFragmenterAmount {
```
tiero, line 1847: In light of #517 we must maybe check the fee account. Depositing low amounts of bitcoins, does not matter anymore (ie. should always work).

The only argument is in favor of not being economically meaningful, but this should be either a warning or a convenience alert in the UI

internal/core/application/operator_service.go
```diff
@@ -1315,7 +1315,40 @@ func (o *operatorService) MarketFragmenterSplitFunds(
...
-	o.splitMarketFragmenterFunds(ctx, mkt, utxos, int(millisatsPerByte), chRes)
+	chRes <- FragmenterSplitFundsReply{
+		Msg: "fetching fee account funds",
+	}
+	feeUtxos, err := o.getAllUnspentsForAccount(ctx, domain.FeeAccount)
+	if err != nil {
+		errMsg := fmt.Errorf(
+			"error while fetching fee account funds: %s", err,
+		)
+		if err == domain.ErrVaultAccountNotFound {
```
tiero, line 1326: maybe warn also fee treshold?

altafan: No. If the fee account is not found it means that no addresses have ever been generated. This means that the operator is trying to deposit into the market (via fragmenter) before funding the fee account.

internal/core/application/operator_service.go
```diff
@@ -1315,7 +1315,40 @@ func (o *operatorService) MarketFragmenterSplitFunds(
...
+					"You need to deposit some LBTC funds to this account used to " +
+					"pay for network fees of transactions",
+			)
+		}
+		chRes <- FragmenterSplitFundsReply{
+			Err: errMsg,
+		}
+		return
+	}
+	if len(feeUtxos) <= 0 {
```
tiero, line 1338: maybe checking the dust limit here, but maybe much  better to use the fee treshold usually a split can be 30k sats

altafan: We check for fee account low balance at every deposit, withdraw or trade transaction.

Here is enough to make sure that the fee account owns at least one utxo because in case the fee account balance is too low, the user should have already been notified (warn on stdout, webhook if registered, and we do not allow amount under 5k sats in fee fragmenter).
Eventually, the fragmentation will return error when crafting the transaction because the of the failing coin selection.

PR #524, opened 2021-12-10
Allow fragmenting only one asset of the pair for already funded markets

### LINKED ISSUE #504: Improve error message with more actionable action the user must take to remediate
Some ideas

- **ErrMarketNotPriced** should be "The selected strategy mandates price to be updated manually"
- **ErrMarketIsClosed** should be  "The market is paused, please open it first"
- **ErrMarketMustBeClosed** should be "The market is active, please pause it first"

### LINKED ISSUE #521: `tdex market fixedfee --base_fee 0 --quote_fee 0` ignores the 0 values
The UpdateMarketFixedFee seems to discard 0 values as if the protobuf request is empty. Those should be instead used as valid values﻿

### LINKED ISSUE #522: Fragmenter Market: It should be possible to deposit a single asset in the pair

### LINKED ISSUE #523: ProposeTrade and TradeComplete permissions error
Calling the ProposeTrade and CompleteTrade it gets 
```
{
  "error": "2 UNKNOWN: /Trade/ProposeTrade: unknown permissions required for method"
}
```﻿

### LINKED ISSUE #525: Trade.txUrl points to api instead of UI explorer

### PR DESCRIPTION
**Pull request by altafan**
Before this, every market fragmentation required to send both assets to the fragmenter account. Now, this behavior is maintained only in case the market has zero balance, otherwise the operator is allowed to fragment funds of only one asset of the pair.

- Closes #522.
- closes #521 by using the special `-1` value to ignore change to base/quote fee instead of using `0`.
- closes #525 by using `localhost:5001` for regtest instead of `localhost:3001`
- closes #523 and added a test to make sure that all methods are either whitelisted or restricted by a permission
- closes #504 by updating the error messages as requested

Please @tiero review this.

### DIFF
--- a/internal/core/application/operator_service.go
+++ b/internal/core/application/operator_service.go
@@ -1792,10 +1792,6 @@ func (o *operatorService) claimDeposit(
 func verifyMarketFunds(
 	market *domain.Market, unspents []domain.Unspent,
 ) error {
-	if !market.IsStrategyBalanced() {
-		return nil
-	}
-
 	outpoints := make([]domain.OutpointWithAsset, 0, len(unspents))
 	for _, u := range unspents {
 		outpoints = append(outpoints, u.ToOutpointWithAsset())
--- a/internal/core/application/operator_service.go
+++ b/internal/core/application/operator_service.go
@@ -2059,24 +2059,25 @@ func (o *operatorService) splitMarketFragmenterFunds(
 	// If the market has zero balance, it's mandatory to send both base and quote
 	// asset to the market fragmenter account. Otherwise, it's possible to split
 	// funds of only one of the assets of the pair.
-	marketBalance, _ := getBalanceForMarket(o.repoManager, ctx, market, false)
-	if market.IsStrategyBalanced() && marketBalance.BaseAmount == 0 &&
-		marketBalance.QuoteAmount == 0 {
-		if assetValuePair.baseValue == 0 {
-			chRes <- FragmenterSplitFundsReply{
-				Err: fmt.Errorf(
-					"missing base asset funds",
-				),
+	if market.IsStrategyBalanced() {
+		marketBalance, _ := getBalanceForMarket(o.repoManager, ctx, market, false)
+		if marketBalance.BaseAmount == 0 && marketBalance.QuoteAmount == 0 {
+			if assetValuePair.baseValue == 0 {
+				chRes <- FragmenterSplitFundsReply{
+					Err: fmt.Errorf(
+						"missing base asset funds",
+					),
+				}
+				return
 			}
-			return
-		}
-		if assetValuePair.quoteValue == 0 {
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by tiero** [comment]
I was thinking that maybe we should allow unbalanced from the start maybe: ie. want to sell NFT or just sell one side of the pair? since we declare the asset hahses now should not be a problem?

Eventually if trader tries to buy the 0 valued side, it will simply get error of not enough liquidity

**Comment by tiero** [comment]
It should be possible to deposit only one asset of the pair via deposit/claim

**Comment by altafan** [comment]
This makes sense. The daemon doesn't support the UNBALANCED market strategy though, which is the strategy that fits this need.

I think the daemon shouldn't behave like that until it supports UNBALANCED strategy.

**Comment by altafan** [comment]
> Well the pluggable already works

Correct.

>  For the BALANCED I understand that 0 can't work for x * y formula. Even if we had the unbalanced already, how would work? The daemon will make BALANCED not activable or simply we will return an error?, ie having a dependency between strategy and balances

It would just be up to the user to change the strategy of the market **BEFORE** sending the splitting the funds sent to the fragmenter. After all, this is what the operator has to do also for a market with pluggable strategy.

I'm going to add a check to "enable" the behavior of the description **ONLY** for markets with balanced strategy. 

All other strategies should always allow single asset fragmentations without such restrictions.

**Comment by tiero** [comment]
> the user to change the strategy of the market BEFORE 

But we should make sure this does not become footgun ie. when we verify the deposit in case of BALANCED strategy, we should alert user with a clear instruction what to do next: change the strategy and claim again should be able to recover for this situation.

Not clear if this is the current behavior

**Comment by altafan** [comment]
@tiero while increasing the verbosity of the error messages, I took the opportunity to address also other issues regarding marginal bugs:
- closes #521 by using the special `-1` value to ignore change to base/quote fee instead of using `0`.
- closes #525 by using `localhost:5001` for regtest instead of `localhost:3001`
- closes #523 and added a test to make sure that all methods are either whitelisted or restricted by a permission
- closes #504 by updating the error messages as requested

File: internal/core/application/operator_service.go [hunk 1]
```diff
@@ -2052,21 +2056,28 @@ func (o *operatorService) splitMarketFragmenterFunds(
...
 
-	if assetValuePair.baseValue == 0 {
-		chRes <- FragmenterSplitFundsReply{
-			Err: fmt.Errorf(
-				"missing base asset funds",
-			),
+	// If the market has zero balance, it's mandatory to send both base and quote
+	// asset to the market fragmenter account. Otherwise, it's possible to split
+	// funds of only one of the assets of the pair.
+	marketBalance, _ := getBalanceForMarket(o.repoManager, ctx, market, false)
```
tiero, line 2062: maybe lets make this query AFTER we know the strategy is balances, to save extra lookup

altafan:
@tiero I pushed a new commit 52dca41 in order to close #514.

PR #529, opened 2021-12-15
Use fee account for marketfragmenter withdrawals

### LINKED ISSUE #528: [MarketFragmenter] Fee account not used for withdrawals
The Fee account should be used  by the MarketfFagmenter to top-up for network fees of withdrawal transactions.

[Here](https://github.com/tdex-network/tdex-daemon/blob/master/internal/core/application/operator_service.go#L1359) passing `nil` instead of the market makes the fragmenter **NOT** using the fee account instead.

altafan:
Closes #528.

Please @tiero, review this.


sekulicd:
LGTM

PR #532, opened 2021-12-15
Allow to withdraw all LBTC funds from Fee account

### LINKED ISSUE #531: [Fee account] Unable to withdraw all funds
It's not possible to withdraw all the LBTC funds from the fee account because it doesn't subtract fees in such case.

The `sendToManyWithoutFeeTopup` function used to craft the transaction has an option `subtractFees` which should be set to true in case the amount matches the fee account balance.

altafan:
Closes #531.

Please @tiero, review this.


sekulicd:
LGTM

PR #535, opened 2021-12-15
[Config] Unset base asset if quote is fixed

### LINKED ISSUE #534: TDEX_QUOTE_ASSET and TDEX_BASE_ASSET should be mutually exclusive
At the moment, the daemon uses a default base asset (LBTC mainnet).

In order to fix a quote asset instead, one must therefore export 2 variables:
```
$ export TDEX_QUOTE_ASSET="myasset"
$ export TDEX_BASE_ASSET="" # to "empty" the default base asset
```

Ideally, if the TDEX_QUOTE_ASSET is set, the default base asset should be automatically set to an empty string instead of forcing the user to do that.

Also, should these 2 config params be mutually exclusive? Starting a daemon with both fixed means starting a daemon that supports only a specific market which is something that may do not make sense.

altafan:
Closes #534.

Please @tiero review this.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -295,18 +295,17 @@ func validate() error {
 		return fmt.Errorf("%s must be equal or greater than 0.1", TradeSatsPerByte)
 	}
 
-	if viper.IsSet(QuoteAssetKey) && viper.IsSet(BaseAssetKey) {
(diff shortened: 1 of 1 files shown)


sekulicd:
LGTM

PR #543, opened 2021-12-21
remoteSigner plugin added to wallet

### LINKED ISSUE #482: Detach Signer to be able to plug external key management components
Let's say we want to have GDK to handle keys and signing or having KMS/Lambda doing this externally etc.. would be better to detach keys from the daemon so we can plug many "signing backend" without embedding in the main binary

sekulicd:
This adds option to plug in different Signers inside Wallet.
If no Signer is provided daemon uses exiting signing wrapped in InnerSigner.

This closes #482 


tiero:
Can we add an integration test with a remote signer in another process? 

It can be `elements-cli` or another `pkg/wallet` but different process

PR #559, opened 2022-03-14
[GetMarketReport] Add volume granularity

Janaka-Steph:
Add volume granularity in order to have a detailed report with all data points, from which we can draw a volume chart.

- GetMarketReport is called with an additional `granularity` parameter of type TimeRange.
- Values of `PredefinedPeriod` in TimeRange have been changed to also work with `granularity` parameter. Remove `LAST` prefix, and add a new FOUR_HOURS period which is a classic period on charting. 

- `MarketReport` returns a new `detail_volume` array of `MarketVolume`. 
- `MarketVolume` get new `start_date` and `end_date` specifying the covered time period. `end_date` could eventually be omitted since we can calculate this information from `granularity`.

Please review @tiero @sekulicd @altafan


api-spec/protobuf/operator.proto
```diff
@@ -753,10 +760,12 @@ message CustomPeriod {
...
-  YEAR_TO_DATE = 5;
-  ALL = 6;
+  HOUR = 1;
+  FOUR_HOURS = 2;
+  DAY = 3;
+  WEEK = 4;
+  MONTH = 5;
+  THREE_MONTHS = 6;
+  YEAR_TO_DATE = 7;
+  ALL = 8;
```
tiero, line 770: nack

api-spec/protobuf/operator.proto
```diff
@@ -710,19 +710,22 @@ message Page {
 }
 
 message GetMarketReportRequest {
-  // The marker for which returning the report of collected fees.
+  // The market for which returning the report of collected fees.
   Market market = 1;
   // time_range fetch market report from time range
   TimeRange time_range = 2;
+  // level of detail for the specified time_range
+  TimeRange granularity = 3;
```
tiero, line 718: Lets specify here a dedicated enum for granularity

api-spec/protobuf/operator.proto
```diff
@@ -755,8 +762,19 @@ enum PredefinedPeriod {
...
+}
+
+enum Granularity {
+  HOUR = 0;
+  FOUR_HOURS = 1;
+  DAY = 2;
+  WEEK = 3;
+  MONTH = 4;
+  THREE_MONTHS = 5;
+  YEAR = 6;
```
tiero, line 779: Does a timeframe of three months or entire here makes sense?

Janaka-Steph: No it doesn't make sense, I will remove.

api-spec/protobuf/operator.proto
```diff
@@ -755,8 +762,17 @@ enum PredefinedPeriod {
   NULL = 0;
   LAST_HOUR = 1;
   LAST_DAY = 2;
-  LAST_MONTH = 3;
-  LAST_3_MONTHS = 4;
-  YEAR_TO_DATE = 5;
-  ALL = 6;
+  LAST_WEEK = 3;
+  LAST_MONTH = 4;
+  LAST_3_MONTHS = 5;
```
altafan, line 767:
```suggestion
  LAST_THREE_MONTHS = 5;
```

api-spec/protobuf/operator.proto
```diff
@@ -755,8 +762,17 @@ enum PredefinedPeriod {
...
+  YEAR_TO_DATE = 6;
```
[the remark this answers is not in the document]
tiero: this is common language in trading charts and so on, it means let me know the current year (ie. 2022) till today

Janaka-Steph:
Let me know if `Granularity` should be renamed `TimeFrame` (proper trading term) or `Candle`.
The idea behind not using `Candle` is that the report could be used for something else than drawing a chart.

tiero (approved):

PR #566, opened 2022-03-17
GetMarketReport RPC volume information

### LINKED ISSUE #556: [GetMarketReport] PredefinedPeriod should include LAST_WEEK

### LINKED ISSUE #558: [GetMarketReport] quoteVolume is equal to baseVolume

### LINKED ISSUE #561: Add `TimeFrame` to calculate fragments of MarketVolume divided by the granularity chosen
As requested here #559  we should calculate the volume for each timeframe given a specific granularity (not only the total one since the beginning)

sekulicd:
This adds add updates to the GetMarketReport RPC as requested by bellow issues.
This updates application tests pkg name as packages should have same name as file/func that is tested.
For the sake of testing GetMarketReport this also adds few mocks impl.

This closes #561 
This closes #558 
This closes #556


tiero (approved):
Beside the pkg rename for testing files (wait for @altafan comment) rest LGTM

internal/core/application/mocks_test.go
```diff
@@ -1,12 +1,11 @@
-package application_test
+package application
```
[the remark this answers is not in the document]
altafan: nACK.

> This updates application tests pkg name as packages should have same name as file/func that is tested.

Tests **must not** have the same package name of the tested one. The usefulness of `_test` suffix is:
1. tests are not imported in case the tested package is required by another one.
2. as being in a different package, the tests have the same view of the tested package as those actually importing it for concrete use. This way only exported types and function can be tested since they will be the only usable.

Please revert the change.

internal/core/application/operator_service.go
```diff
@@ -1656,21 +1664,67 @@ func (o *operatorService) GetMarketReport(
...
+	for {
+		if end.Equal(start) || end.Before(start) {
+			return groupedVolume
+		} else {
+			groupedVolume = append(groupedVolume, MarketVolume{
+				StartTime: end.Add(-time.Hour * time.Duration(groupByHours)),
+				EndTime:   end,
+			})
+			end = end.Add(-time.Hour * time.Duration(groupByHours))
+		}
```
altafan, line 1724:
```suggestion
		} else {
		        next := end.Add(-time.Hour * time.Duration(groupByHours))
			groupedVolume = append(groupedVolume, MarketVolume{
				StartTime: next,
				EndTime:   end,
			})
			end = next
		}
```

internal/core/application/operator_service.go
```diff
@@ -1656,21 +1664,67 @@ func (o *operatorService) GetMarketReport(
...
+			StartTime:   startTime,
+			EndTime:     endTime,
 		},
+		GroupedVolume: groupedVolume,
 	}, nil
 }
 
+// initGroupedVolume initialised list of MarketVolume with grouped time slots(start/end date)
+//ordered desc
+func initGroupedVolume(start, end time.Time, groupByHours int) []MarketVolume {
```
altafan, line 1713:
```suggestion
// initGroupedVolume splits the given time range (start, end) into a list of MarketVolume, ie. smaller consecutive time ranges of numHours hours in descending order.
// Example:
// in: 2009-11-10 19:00:00 (start), 2009-11-11 00:00:00 (end), 2 (numHours)
// out: [
//   {end: 2009-11-11 00:00:00, start: 2009-11-10 22:00:01},
//   {end: 2009-11-11 22:00:00, start: 2009-11-10 20:00:01},
//   {end: 2009-11-10 20:00:00, start: 2009-11-10 19:00:00},
// ]
func initGroupedVolume(start, end time.Time, numHours int) []MarketVolume {
```

internal/core/application/operator_service.go
```diff
@@ -1656,21 +1664,67 @@ func (o *operatorService) GetMarketReport(
...
+		if end.Equal(start) || end.Before(start) {
+			return groupedVolume
+		} else {
+			groupedVolume = append(groupedVolume, MarketVolume{
+				StartTime: end.Add(-time.Hour * time.Duration(groupByHours)),
+				EndTime:   end,
+			})
+			end = end.Add(-time.Hour * time.Duration(groupByHours))
+		}
+	}
```
altafan, line 1725: we must add here just a second more to avoid collisions between `MarketVolume[i].Start` and `MarketVolume[i+1].End`, which would cause, then, a trade with exactly that timestamp to be repeated in 2 different slots.
The very last slot doesn't need this extra second, but its start time must be exactly the timerange's start time:
```suggestion
		} else {
		        nextEnd := end.Add(-time.Hour * time.Duration(groupByHours))
		        nextStart := start
		        if nextEnd.Sub(start).Seconds() > 0 {
		            nextStart = nextEnd.Add(time.Second)
		        }
			groupedVolume = append(groupedVolume, MarketVolume{
				StartTime: nextStart,
				EndTime:   end,
			})
			end = nextEnd
		}
	}
```

internal/core/application/mocks_test.go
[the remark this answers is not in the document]
sekulicd: I see there are people in community that are using both "white-box" and "black-box" testing techniques, one such discussion can be found in [here](https://stackoverflow.com/questions/19998250/proper-package-naming-for-testing-with-the-go-language).
"Black box" technique is used, as u said, to prevent circular dependencies and to test only exported stuff.
We should not have issues with circular dependencies as our layers are divided vertically, meaning application can import lower domain level and not vice verse, and imported pkg's do not import any of our pkg's.
Other reason, "to test only exported stuff", seems to me as disadvantage in out case since we want to do unit testing and test both exported and unexported funcs like in my case initGroupedVolume. 
I see that in "black box" testing it is recommended to use additional file with suffix *_internal_test.go which would have same pkg name as unexported func being tests.

Anyways, my primary motivation, for updating test pkg name, was to test initGroupedVolume func, so how do you want to proceed with it? to leave changes i did (update test pkg name) or to create additional test file for this fucn with suffix *_internal_test.go ? @tiero @altafan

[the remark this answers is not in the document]
tiero: If we are renaming current test suites for this only function, I would go with `*_internal_test.go` approach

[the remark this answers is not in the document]
altafan: I understand that both "white" and "black" box testing have their pros and cons, but in general, I think this change is something that requires a proper flow to be eventually committed, therefore opening an issue, and eventually proceeding with changes AFTER a discussion.

I don't like the idea of this PR of having the application layer white-box tested, while the rest of the tests are ALL black-box. If the change is done, it must be done for ALL packages.

Another thing I don't like so much right now is having utils function sparse in the application layer. In fact, You can see `repo_utils.go` and `explorer_utils.go` as attempts  of grouping them into one or more util files.

My proposal is: what if we pack util functions altogether (or divided in more `xxx_util.go` in a way that makes sense)  and we "white-box" test only this/these files, leaving the rest "black-box" as already done?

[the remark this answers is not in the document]
sekulicd: > I understand that both "white" and "black" box testing have their pros and cons, but in general, I think this change is something that requires a proper flow to be eventually committed, therefore opening an issue, and eventually proceeding with changes AFTER a discussion.

I don't like the idea of this PR of having the application layer white-box tested, while the rest of the tests are ALL black-box. If the change is done, it must be done for ALL packages.

Agree.

> Another thing I don't like so much right now is having utils function sparse in the application layer. 

Dont agree with this one, from my pov it is ok to have private methods/functions inside application service files, this encourages splitting complex service methods to more granular funcs which improves testability and readability.
Utils methods should be like pkg's, something could be reusable in various places.

internal/core/application/operator_service.go

internal/core/application/operator_service.go

sekulicd:
@tiero @altafan test pkg name reverted.

internal/core/application/operator_service_internal_test.go
```diff
@@ -0,0 +1,47 @@
+package application
```
altafan, line 1:
```suggestion
package application_internal_test
```

internal/core/application/operator_service_internal_test.go
```diff
@@ -0,0 +1,47 @@
...
+		{
+			name: "2",
+			args: args{
+				start:            start,
+				end:              start.Add(12 * time.Hour),
+				groupByTimeFrame: 1,
+			},
+			want: 12,
+		},
+	}
```
altafan, line 38: I think it would be really valuable to have a series of test cases here with `start`, `end` assuming values reflecting the `TimeRange` proto enum entries and for each of these ranges, `groupByTimeFrame` assuming all the values reflecting `TimeFrame` proto enum.

I guess some of them should be invalid: in general `groupByTimeFrame` shouldn't be <= `TimeRange.EndTime - TimeRange.StartTime`?

I mean, it would not have that much sense for example a timerange of a week, with a timeframe of a month or a year, do you agree?

tiero: timeframe < timerange ofc

if timeframe == timerange you would see a single candle in the chart so to speak, not that useful, but valid

altafan (changes requested):
in application tests, we don't want to use mocked repositories. In that layer, we have so-called "component tests" testing the flow of services using "real" dependencies (except for the mocked-explorer).

Of course, I understand the importance of a unit test for GetMarketReport, therefore it's ok to have one, but for now, to stick with all other tests of this package, you should create a new `newOperatorServiceForMarketReport()` function where you can add to a real repo manager all the data you need for your unit test (just like you've done with the mocked repo).

Since with this PR we unexpectedly introduced tests for internal methods of the application package, I will open a ticket to make a new proposal in order to have black-box component tests, and white-box unit tests.

For the moment, please

internal/core/application/operator_service_internal_test.go
sekulicd: Done

sekulicd:
@tiero @altafan removed mocks, updated test case.

Janaka-Steph:
Screenshot 2022-03-18 at 12 55 54

In detailVolumeList, what about only returning slots with positive volume ?
Currently it returns all slots, even if zero volume. Technically valid but seems a waste of bandwith.

Date seems to be either not valid or overly complicated.
Should be like this `2019-10-12T07:20:50.52Z` but we have `2021-01-04 11:53:51.055556844 +0000 UTC m=-37846485.410451690`
This is a different format, we don't have separator `T` and `Z`.
https://medium.com/easyread/understanding-about-rfc-3339-for-datetime-formatting-in-software-engineering-940aa5d5f68a

Janaka-Steph:
Requesting `PredefinedPeriod.ALL` returns data from now until `2021-01-01 00:00:00`. Why this date?

tiero:
> Requesting `PredefinedPeriod.ALL` returns data from now until `2021-01-01 00:00:00`. Why this date?

There was no usage of tdex daemons in production before that date anyway?

sekulicd:
> Screenshot 2022-03-18 at 12 55 54
> 
> In detailVolumeList, what about only returning slots with positive volume ? Currently it returns all slots, even if zero volume. Technically valid but seems a waste of bandwith.
> 
> Date seems to be either not valid or overly complicated. Should be like this `2019-10-12T07:20:50.52Z` but we have `2021-01-04 11:53:51.055556844 +0000 UTC m=-37846485.410451690` This is a different format, we don't have separator `T` and `Z`. https://medium.com/easyread/understanding-about-rfc-3339-for-datetime-formatting-in-software-engineering-940aa5d5f68a
> 
> `time.Time` used in the code returns the Golang format https://pkg.go.dev/time#Time Seems it needs this https://pkg.go.dev/time#Time.MarshalJSON or https://pkg.go.dev/time#Time.MarshalText to convert to RCF3339 ? The doc of `Time` func says `The returned string is meant for debugging; for a stable serialized representation, use t.MarshalText, t.MarshalBinary, or t.Format with an explicit format string.` https://pkg.go.dev/time#Time.String

@tiero what about this? should we filter out slots where base/quote are 0  ?

tiero:
> @tiero what about this? should we filter out slots where base/quote are 0 ?

I think is a good suggestion, there is no point in showing candles in a chart with volume 0. If a timeframe is not shown in the result, client can infer that there has been 0 volume for that time frame and draw the conclusion.

internal/core/application/operator_service.go
```diff
@@ -1656,21 +1664,79 @@ func (o *operatorService) GetMarketReport(
...
+					if swapRequest.GetAssetR() == market.QuoteAsset {
+						volumeBaseAmount = swapRequest.GetAmountP() + v.BaseVolume
+						volumeQuoteAmount = swapRequest.GetAmountR() + v.QuoteVolume
+					}
+
+					groupedVolume[i] = MarketVolume{
+						BaseVolume:  volumeBaseAmount,
+						QuoteVolume: volumeQuoteAmount,
+						StartTime:   v.StartTime,
+						EndTime:     v.EndTime,
```
tiero, line 1687: shall we [Time.MarshalText](https://pkg.go.dev/time#Time.MarshalText) as suggested by @Janaka-Steph  to "beautify" the string?

internal/core/application/operator_service.go
```diff
@@ -1656,21 +1664,79 @@ func (o *operatorService) GetMarketReport(
...
 			QuoteAmount: uint64(totalFees[market.QuoteAsset]),
+			StartTime:   startTime,
+			EndTime:     endTime,
 		},
 		Volume: MarketVolume{
 			BaseVolume:  uint64(volume[market.BaseAsset]),
-			QuoteVolume: uint64(volume[market.BaseAsset]),
+			QuoteVolume: uint64(volume[market.QuoteAsset]),
+			StartTime:   startTime,
+			EndTime:     endTime,
```
tiero, line 1705: shall we [Time.MarshalText](https://pkg.go.dev/time#Time.MarshalText) as suggested by @Janaka-Steph  to "beautify" the string?

sekulicd: Done

sekulicd:
> > @tiero what about this? should we filter out slots where base/quote are 0 ?
> 
> I think is a good suggestion, there is no point in showing candles in a chart with volume 0. If a timeframe is not shown in the result, client can infer that there has been 0 volume for that time frame and draw the conclusion.

Done

Janaka-Steph:
@tiero @sekulicd `detailVolumeList` now always returns only one element!

internal/core/application/operator_service.go
```diff
@@ -1680,17 +1681,24 @@ func (o *operatorService) GetMarketReport(
...
 				}
 			}
 		}
 	}
 
+	groupedVolume := make([]MarketVolume, 0)
+	for k, _ := range groupedVolumeIndexes {
+		groupedVolume = append(groupedVolume, groupedVolumeTmp[k])
+	}
+
```
tiero, line 1701: Is this piece of code skipping the 0-ed values?

sekulicd: yes

tiero: Is this related https://github.com/tdex-network/tdex-daemon/pull/566#issuecomment-1072573390

sekulicd: I dont think so since there is a [test case](https://github.com/sekulicd/tdex-daemon/blob/cc2b9a9edc5fac90c9c0fe07f986506494b6d001/internal/core/application/operator_service_test.go#L257) that check length and test is passing, this is on application layer and in interface layer we are looping through array in [here](https://github.com/sekulicd/tdex-daemon/blob/cc2b9a9edc5fac90c9c0fe07f986506494b6d001/internal/interfaces/grpc/handler/operator.go#L1298) so i dont see issue in this moment.

sekulicd: @Janaka-Steph are u sure u created more trades in different time slots(candle's) and that you passed correct timeRange/timeFrame args?

tiero:
> @tiero @sekulicd `detailVolumeList` now always returns only one element equals to `totalVolume`!

Can you post the ListTrades result and tell us the payload you used?

altafan:
> I think is a good suggestion, there is no point in showing candles in a chart with volume 0. If a timeframe is not shown in the result, client can infer that there has been 0 volume for that time frame and draw the conclusion.

This is good for bandwidth, true, but the only way for the client to infer that those time frames are 0 is by recomputing the same frames as done by the daemon.

Although the volumes are 0, it's good for the client to receive all of them for building the diagram.
If the daemon sends only the frames with non-0-volumes, the client might struggle to represent them in the diagram because all others 0-volume candles are missing?

Janaka-Steph:
Yes I agree with @altafan actually. Client needs an array with all the dates at least. Recomputing the dates can be prone to errors.

Janaka-Steph:
> > @tiero @sekulicd `detailVolumeList` now always returns only one element equals to `totalVolume`!
> 
> Can you post the ListTrades result and tell us the payload you used for this request?

```
const { data: marketReport } = useGetMarketReportQuery(
    {
      market: marketInfo?.market,
      timeRange: { predefinedPeriod: PredefinedPeriod.ALL },
    },
    { skip: !marketInfo?.market }
  );
```
```
{
    "totalCollectedFees": {
        "baseAmount": 125199,
        "quoteAmount": 0
    },
    "totalVolume": {
        "baseVolume": 50079953,
        "quoteVolume": 1988248936316,
        "startDate": "2021-01-01T00:00:00Z",
        "endDate": "2022-03-18T16:26:53Z"
    },
    "detailVolumeList": [
        {
            "baseVolume": 50079953,
            "quoteVolume": 1988248936316,
            "startDate": "2022-03-18T15:26:54Z",
            "endDate": "2022-03-18T16:26:53Z"
        }
    ]
}
```

sekulicd:
Here are extracted complete_time_unix from trades list:
```
1647620412 Friday, March 18, 2022 4:20:12 PM
1647620320 Friday, March 18, 2022 4:18:40 PM
1647620262 Friday, March 18, 2022 4:17:42 PM
1647620193 Friday, March 18, 2022 4:16:33 PM
```
All are within one hour and based on request:
```
{
  "market": {
    "base_asset": "5ac9f65c0efcc4775e0baec4ec03abdde22473cd3cf33c0419ca290e0751b225",
    "quote_asset": "8656dfbbba1ad643c47a208a89fe0724dbc521bf3e05e641ba0e924665dc3c81"
  },
  "time_range": {
    "predefined_period": "ALL"
  }
}
```
time_zone was not passed and it is "HOUR" by default.

Response looks fine.

sekulicd:
> > I think is a good suggestion, there is no point in showing candles in a chart with volume 0. If a timeframe is not shown in the result, client can infer that there has been 0 volume for that time frame and draw the conclusion.
> 
> This is good for bandwidth, true, but the only way for the client to infer that those time frames are 0 is by recomputing the same frames as done by the daemon.
> 
> Although the volumes are 0, it's good for the client to receive all of them for building the diagram. If the daemon sends only the frames with non-0-volumes, the client might struggle to represent them in the diagram because all others 0-volume candles are missing?

@tiero @Janaka-Steph should we return slots like before?

Janaka-Steph:
> > > I think is a good suggestion, there is no point in showing candles in a chart with volume 0. If a timeframe is not shown in the result, client can infer that there has been 0 volume for that time frame and draw the conclusion.
> > 
> > 
> > This is good for bandwidth, true, but the only way for the client to infer that those time frames are 0 is by recomputing the same frames as done by the daemon.
> > Although the volumes are 0, it's good for the client to receive all of them for building the diagram. If the daemon sends only the frames with non-0-volumes, the client might struggle to represent them in the diagram because all others 0-volume candles are missing?
> 
> @tiero @Janaka-Steph should we return slots like before?

Actually I think it's safer

tiero:
> @tiero @Janaka-Steph should we return slots like before?

Yes, best to do like you did it first.

sekulicd:
> Yes, best to do like you did it first.

Done

PR #574, opened 2022-03-23
Unlock wallet on startup

### LINKED ISSUE #557: Move `unlockerd` inside `tdexd` and add `UNLOCK_PASSWORD_FILE` config ENV VAR option to make unlocker feature inside `tdexd`
TODO:

- [x] move unlocker functionalities inside `tdexd`
- [x] add new ENV VAR
- [ ] change tdex-box docker-compose and mount the volume with pwd.txt inside according to new ENV VAR
- [ ] change documentation in the Dev Portal
- [ ] change documentation in the README

sekulicd:
This adds unlocking wallet on startup if wallet unlock password file is provided through env var.

This closes #557 

### DIFF
--- a/internal/interfaces/grpc/service.go
+++ b/internal/interfaces/grpc/service.go
@@ -1,6 +1,7 @@
 package grpcinterface
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io/ioutil"
@@ -8,7 +9,6 @@ import (
 	"net"
 	"net/http"
 	"path/filepath"
-	"strings"
 
 	log "github.com/sirupsen/logrus"
 	"github.com/soheilhy/cmux"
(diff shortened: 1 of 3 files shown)


internal/interfaces/grpc/service.go
```diff
@@ -220,11 +230,22 @@ func NewService(opts ServiceOpts) (interfaces.Service, error) {
...
 	}
 
+	var walletPassword string
+	if opts.WalletUnlockPasswordFile != "" {
+		walletPasswordBytes, err := ioutil.ReadFile(opts.WalletUnlockPasswordFile)
+		if err != nil {
+			return nil, err
+		}
+
+		walletPassword = strings.TrimSpace(string(walletPasswordBytes))
```
altafan, line 240: the unlockerd service uses:
```
 bytes.TrimFunc(walletPasswordBytes, func(r rune) bool {
	return r == 10 || r == 32
}), nil
```
this function trims both spaces and end-of-lines. I would keep using this trim function.

tiero:
not sure if [this](https://github.com/tdex-network/tdex-daemon/runs/5663868067) is valuable, but looks like we can expose the password on an error?

internal/interfaces/grpc/service.go
sekulicd: Done

PR #575, opened 2022-03-24
Trade grpc-gateway support

### LINKED ISSUE #545: [grpc-gateway] Support `application/json` content-type for Trade interface

sekulicd:
This adds grpc-gateway support for trade interface.

This closes #545 

### DIFF
--- a/internal/interfaces/grpc/http_utils.go
+++ b/internal/interfaces/grpc/http_utils.go
@@ -12,6 +12,7 @@ import (
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"encoding/pem"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net"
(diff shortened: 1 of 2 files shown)


tiero:
Can you test this with a tor hidden service before merging? (can be done easily on your machine)

internal/interfaces/grpc/http_utils.go
```diff
@@ -334,3 +331,32 @@ func isValidGrpcWebOptionRequest(req *http.Request) bool {
 		strings.Contains(accessControlHeader, "x-grpc-web") &&
 		strings.Contains(accessControlHeader, "content-type")
 }
+
+func isHttpRequest(req *http.Request) bool {
+	return strings.ToLower(req.Method) == "get" ||
```
altafan, line 336: we should check also for `post` methods.

internal/interfaces/grpc/service.go
```diff
@@ -489,3 +502,29 @@ func (s *service) startListeningToReadyChan() {
...
+	if tlsKey != "" {
+		tlsConfig, err := getTlsConfig(tlsKey, tlsCert)
+		if err != nil {
+			return nil, err
+		}
+
+		creds = grpc.WithTransportCredentials(credentials.NewTLS(tlsConfig))
+	}
+
+	conn, err := grpc.DialContext(ctx, s.opts.TradeAddress, []grpc.DialOption{creds}...)
```
altafan, line 517: DialContext should be used in combo with a context-with-deadline. Why not using just Dial() here without any ctx?

internal/interfaces/grpc/http_utils.go
sekulicd: Done

internal/interfaces/grpc/service.go
sekulicd: Done

internal/interfaces/grpc/http_utils.go
tiero: mmm gRPC calls are post calls. also we always check for content type json OR get in other proj

tiero: i think this was fine

sekulicd: Done

altafan (commented):
utACK

PR #581, opened 2022-03-30
Permission validation check

### LINKED ISSUE #578: permissions guard prevent calling  Wallet Unlocker
With master 

```
[tdex] rpc error: code = Unknown desc = /tdex.daemon.v1.WalletUnlocker/GenSeed: unknown permissions required for method
```

sekulicd:
This adds validation for RPC's permissions that is checked on startup and in test case.
If new method is added in proto's and if it is not added in permissions.go daemon will not start.
This prevents forgetting adding of new rpc methods to permissions map.  

This closes #578 

### DIFF
--- a/internal/interfaces/grpc/permissions/permissions.go
+++ b/internal/interfaces/grpc/permissions/permissions.go
@@ -1,6 +1,13 @@
 package permissions
 
 import (
+	"fmt"
+
+	log "github.com/sirupsen/logrus"
+
+	tdexv1 "github.com/tdex-network/tdex-daemon/api-spec/protobuf/gen/go/tdex/v1"
+
+	daemonv1 "github.com/tdex-network/tdex-daemon/api-spec/protobuf/gen/go/tdex-daemon/v1"
 	"gopkg.in/macaroon-bakery.v2/bakery"
 )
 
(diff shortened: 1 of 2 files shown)


tiero:
`TestRestrictedMethods` need to be updated?

altafan:
Actually, the tests inside `permissions_test.go` should already accomplish this task.
I mean, [TestRestrictedMethods](https://github.com/tdex-network/tdex-daemon/blob/master/internal/interfaces/grpc/permissions/permissions_test.go#L13) and [TestWhitelistedMethods](https://github.com/tdex-network/tdex-daemon/blob/master/internal/interfaces/grpc/permissions/permissions_test.go#L29) are there to ensure basically that all the methods are either restricted or whitelisted.

Maybe we don't really need the `validatePermissions()` func and the white-box testing here, but rather to just update the black-box tests after the changes we affecting the methods' full paths.

sekulicd:
> Actually, the tests inside `permissions_test.go` should already accomplish this task. I mean, [TestRestrictedMethods](https://github.com/tdex-network/tdex-daemon/blob/master/internal/interfaces/grpc/permissions/permissions_test.go#L13) and [TestWhitelistedMethods](https://github.com/tdex-network/tdex-daemon/blob/master/internal/interfaces/grpc/permissions/permissions_test.go#L29) are there to ensure basically that all the methods are either restricted or whitelisted.
> 
> 
> Actually, the tests inside permissions_test.go should already accomplish this task.
I mean, [TestRestrictedMethods](https://github.com/tdex-network/tdex-daemon/blob/master/internal/interfaces/grpc/permissions/permissions_test.go#L13) and [TestWhitelistedMethods](https://github.com/tdex-network/tdex-daemon/blob/master/internal/interfaces/grpc/permissions/permissions_test.go#L29) are there to ensure basically that all the methods are either restricted or whitelisted.
Maybe we don't really need the `validatePermissions()` func and the white-box testing here, but rather we just need to update the black-box tests after the changes affecting the methods' full paths.

I think its really good to have validatePermissions run inside init method so we detect this issue early and done see harm with testing it in white_box manner.

altafan:
Agreed, the only thing I don't really like is validating the permissions within `init()`.
The downside of this is that we halt the execution of the program directly with a `log.Fatal()`, while it should be better to propagate the error and handle it where the service is started (ie. in `cmd/tdexd.go`) instead.
This way, we can safely stop/close any other service that actually needs to (repo manager, for example).

One solution could be calling `validatePermissions()` within [`NewService()`](https://github.com/sekulicd/tdex-daemon/blob/permission_validation/internal/interfaces/grpc/service.go#L228) only when macaroons are enabled.

@sekulicd @tiero What do you think?

tiero:
Yep init "hides" too much, if not that much polluting, let's move it to NewService

sekulicd:
> Yep init "hides" too much, if not that much polluting, let's move it to NewService

Done

PR #585, opened 2022-04-05
Add backward compatibility with old version of tdex-protobuf

### LINKED ISSUE #583: Serve both trade.proto api specs (matser & v1) at same time to allow smooth upgrade for clients to new protos being generate

altafan:
This adds support for the old version of tdex-protobuf (before migrating to buf):
- adds a new grpc tradeOldHandler using the same application service, but depending on tdex-protobuf#master instead of #v1 
- adds registration of the handler above to the trade grpc server
- adds macaroon permissions to whitelist accordingly

NOTE: the old version of tdex-protobuf does not have support for grpc-gateway, therefore that part has not been affected by changes.

Closes #583.

Please @tiero review this.


sekulicd:
LGTM

altafan:
@tiero I pushed a `v0.8.8-rc.0` and updated to this version the testnet providers.
You can test on your own that the daemons are able to serve requests for both versions of tdex-protobuf (legacy and v1).

PR #589, opened 2022-04-12
Lint protos

### LINKED ISSUE #586: Lint protos
Protos are not linted, nor the CI ensures this condition. We must add a `proto-lint` target to the make file, a wrapper for `buf lint`, and add also a step to the gh action to make sure protos are linted for both pull requests and releases.

altafan:
This updates the definition of the protos in order to be compliant with buf's linter.

This also adds a new `proto-lint` target to the Makefile used by the `proto` target when compiling the stubs to prior ensure that they are listed.

This also updates the gh actions, in particular `ci.yml` and `release.yml` so that they make sure the protos are linted before doing their usual job.

Closes #586.

Please @tiero review this.


.github/workflows/ci.yml
```diff
@@ -11,28 +11,34 @@ jobs:
...
+          input: 'api-spec/protobuf'
 
       - name: Run Gosec Security Scanner
-        run: |
-          export PATH=$PATH:$(go env GOPATH)/bin
-          curl -sfL https://raw.githubusercontent.com/securego/gosec/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.9.3
-          gosec -severity high -quiet ./...
+        uses: securego/gosec@master
+        with:
+          args: '-severity high -quiet ./...'
```
tiero, line 33: we used the other way since wasn't not working somehow (ie. reporting the security high)

PR #597, opened 2022-04-25
Buf: Update paths to more idiomatic folder locations

tiero:
TL;DR

- Dropped `go` folder inside `gen`
- Moved swagger stuff inside same folder with other stubs under `gen`


sekulicd (commented):
From my POV, swagger.json is api spec and maybe has more sense to be located with other spec files(protobuf) or maybe to some other location like /api-spec/swagger, then in same location as go generated stubs, beside that ACK.

Makefile
```diff
@@ -22,7 +22,7 @@ build-tdexdconnect:
 proto: proto-lint
```
sekulicd, line 22: We should add external dep download cmd's (buf mod update & buf build), also find it useful to run go mod tidy in the end, so my suggestion would be:

```
## proto: compile proto files
proto: proto-lint
	@echo "Compiling stubs..."
	@cd api-spec/protobuf; buf mod update; buf build
	@buf generate buf.build/tdex-network/tdex-protobuf
	@buf generate
	@make install
```

altafan (approved):

sekulicd (approved):
ACK

PR #600, opened 2022-05-02
Restore watch for trades and unspents after restart

### LINKED ISSUE #564: Start goroutines for unlocking utxos after a crash/restart

altafan:
This makes the trade service restore the watch for accepted trades and locked utxos.

NOTE: if a trade is already expired, the service doesn't update it's status immediately and releases the lock of the utxos, but it waits for a minute before doing that. This to give the time to the bc listener to eventually be notified about a settled trade so that it can correctly update its status and spend the utxos instead.

Closes #564.

Please @tiero review this.

### DIFF
--- a/internal/core/application/trade_service.go
+++ b/internal/core/application/trade_service.go
@@ -94,7 +94,7 @@ func newTradeService(
 	net *network.Network,
 	feeAccountBalanceThreshold uint64,
 ) *tradeService {
-	return &tradeService{
+	svc := &tradeService{
 		repoManager:                repoManager,
 		explorerSvc:                explorerSvc,
 		blockchainListener:         bcListener,
@@ -105,6 +105,37 @@ func newTradeService(
 		feeAccountBalanceThreshold: feeAccountBalanceThreshold,
 		lock:                       &sync.Mutex{},
 	}
+
+	// In case of restart, restore watching trades/outpoints: start observing the
+	// inputs of each trade even if it's already expired.
+	// The idea is to wait at least for a minute to give the time to the bc
+	// listener to make sure the trade has been settled.
+	// Otherwise, the trade will expire and the utxos will be unlocked.
+	trades, _ := svc.repoManager.TradeRepository().GetAllTrades(context.Background())
+	for i := range trades {
+		trade := trades[i]
+		if !trade.Status.Failed && trade.IsAccepted() {
+			ptx, _ := pset.NewPsetFromBase64(trade.PsetBase64)
+			utxos := make([]explorer.Utxo, 0, len(ptx.Inputs))
+			for _, in := range ptx.UnsignedTx.Inputs {
+				utxos = append(utxos, utxoKey{bufferutil.TxIDFromBytes(in.Hash), in.Index})
+			}
+
+			svc.blockchainListener.StartObserveOutpoints(utxos, trade.ID.String())
+			waitingTime := time.Minute
+
+			if !trade.IsExpired() {
+				waitingTime += time.Duration(trade.ExpiryTime-uint64(time.Now().Unix())) * time.Second
+			}
+			log.Debugf("trade with id %s will expire in ~%s", trade.ID, waitingTime)
+			go func() {
+				time.Sleep(waitingTime)
+				svc.checkTradeExpiration(trade.ID, utxos)
+			}()
+		}
+	}
+
+	return svc
 }
 
 func (t *tradeService) GetTradableMarkets(ctx context.Context) (
@@ -381,10 +412,12 @@ end:
 		)
 		log.Debugf("locked %d unspents", lockedUnspents)
 
+		// admit a tollerance of 1 minute past the expiration time.
+		waitingTime := t.expiryDuration + time.Minute
+		log.Debugf("trade with id %s will expire in ~%s", trade.ID, waitingTime)
 		// set timer for trade expiration
 		go func() {
-			// admit a tollerance of 1 minute past the expiration time.
-			time.Sleep(t.expiryDuration + time.Minute)
+			time.Sleep(waitingTime)
 			t.checkTradeExpiration(trade.ID, fillProposalResult.SelectedUnspents)
 		}()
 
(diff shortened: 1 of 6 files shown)


internal/infrastructure/storage/db/badger/trade_repository_impl.go
```diff
@@ -210,34 +210,45 @@ func (t tradeRepositoryImpl) getTrade(
...
 	ID uuid.UUID,
 ) (*domain.Trade, error) {
-	var trade domain.Trade
+	var trades []domain.Trade
 	var err error
+	query := badgerhold.Where("ID").Eq(ID)
 	if ctx.Value("tx") != nil {
 		tx := ctx.Value("tx").(*badger.Txn)
-		err = t.store.TxGet(tx, ID, &trade)
+		err = t.store.TxFind(tx, &trades, query)
```
tiero, line 218: whats difference between Find & Get?

I recall some badger bug or something

internal/infrastructure/storage/db/badger/trade_repository_impl.go
```diff
@@ -210,34 +210,45 @@ func (t tradeRepositoryImpl) getTrade(
...
 	ctx context.Context,
 	ID uuid.UUID,
 	trade domain.Trade,
 ) error {
+	query := badgerhold.Where("ID").Eq(ID)
 	if ctx.Value("tx") != nil {
 		tx := ctx.Value("tx").(*badger.Txn)
-		return t.store.TxUpdate(tx, ID, trade)
+		return t.store.TxUpdateMatching(tx, &domain.Trade{}, query, func(record interface{}) error {
+			tr := record.(*domain.Trade)
```
tiero, line 242: seem panic prone. shall we have a domain method?

internal/infrastructure/storage/db/badger/trade_repository_impl.go
altafan: With Find we use a sql-like query to retrieve an element, while Get returns the value that matches the given key. I didn't go in deep but it's likely that Find returns a value that matches the given condition without checking the key like Get does instead.

The real bug here is using a uuid.UUID as key instead of a built-in type like string. This update prevents very weird behaviors by the repository btw. (Un?)Luckily for us, no providers have ever suffered from this.

internal/infrastructure/storage/db/badger/trade_repository_impl.go
altafan: no this is not panic prone, the 2 argument of the function (`&domain.Trade{}`) defines the concrete type of `record`, therefore it's safe to cast without checking the type. We already do the same in the unspent repo btw.

> shall we have a domain method?

No, because this is related to the badger impl. The domain must not depend on or know anything about badger or any other concrete implementation.

internal/infrastructure/storage/db/badger/trade_repository_impl.go
sekulicd: nACK
Get is much faster, since it is searching by key, i got bellow benchmarks:
WithFind            1468791346 ns/op
WithGet		  66934           ns/op

internal/core/application/trade_service.go
```diff
@@ -105,6 +108,40 @@ func newTradeService(
...
+	// In case of restart, restore watching trades/outpoints: start observing the
+	// inputs of each trade even if it's already expired.
+	// The idea is to wait at least for a minute to give the time to the bc
+	// listener to make sure the trade has been settled.
+	// Otherwise, the trade will expire and the utxos will be unlocked.
+	trades, _ := svc.repoManager.TradeRepository().GetAllTrades(context.Background())
+	for i := range trades {
+		trade := trades[i]
+		if !trade.Status.Failed && trade.IsAccepted() {
+			ptx, _ := pset.NewPsetFromBase64(trade.PsetBase64)
```
sekulicd, line 124: should we put all from L124 - L140 in another goroutine? otherwise we need to wait atleast one min for each accepted trade.

internal/infrastructure/storage/db/badger/trade_repository_impl.go
tiero: may or may not be related https://github.com/tdex-network/tdex-daemon/pull/454

internal/infrastructure/storage/db/badger/trade_repository_impl.go
```diff
@@ -210,34 +210,45 @@ func (t tradeRepositoryImpl) getTrade(
...
+			*tr = trade
+			return nil
+		})
 	}
-	return t.store.Update(ID, trade)
+	return t.store.UpdateMatching(&domain.Trade{}, query, func(record interface{}) error {
```
sekulicd, line 247: This also looks like potential performance setback.(again this difference between searching row by key or query)

altafan: This will be fixed in the PR containing the wallet detachment. For now it's needed because of the unfortunate choice of using the entire struct uuid.UUID as key for a stored trade.
Without this, it might happen that the daemon duplicates the same trade which is a serious bug that only for (un)luck has never been caught.

internal/infrastructure/storage/db/badger/trade_repository_impl.go
altafan: Same here: this will be fixed, ie. restored to use Get, in next PR, but for now it's need to prevent unexpected behaviors.

internal/infrastructure/storage/db/badger/trade_repository_impl.go
tiero: Add a comment about this in the code please

internal/infrastructure/storage/db/badger/trade_repository_impl.go
tiero: add a comment?

altafan: Done. Sorry i completely miss this 🙏

tiero (approved):

PR #617, opened 2022-08-17
Run operator and trade services on same port

### LINKED ISSUE #613: Allow to bind operator and trade interface on the same port

sekulicd:
This adds option to run trade and operator services on same port if defined by config.

This closes #613 

@tiero please review.

### DIFF
--- a/cmd/tdexd/main.go
+++ b/cmd/tdexd/main.go
@@ -59,6 +59,7 @@ var (
 	rescanRangeStart              = config.GetInt(config.RescanRangeStartKey)
 	rescanGapLimit                = config.GetInt(config.RescanGapLimitKey)
 	walletUnlockPasswordFile      = config.GetString(config.WalletUnlockPasswordFile)
+	runOnOnePort                  = config.GetBool(config.RunOnOnePortKey)
 
 	version = "dev"
 	commit  = "none"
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -95,6 +95,8 @@ const (
 	//password for unlocking the wallet, if provided wallet will be unlocked
 	//automatically
 	WalletUnlockPasswordFile = "WALLET_UNLOCK_PASSWORD_FILE"
+	// RunOnOnePortKey is used to run the daemon on one port as requested by some users
```
tiero, line 98:
```suggestion
	// RunOnOnePortKey is used to run the daemon on one port 
```

cmd/tdexd/main.go
```diff
@@ -59,6 +59,7 @@ var (
 	rescanRangeStart              = config.GetInt(config.RescanRangeStartKey)
 	rescanGapLimit                = config.GetInt(config.RescanGapLimitKey)
 	walletUnlockPasswordFile      = config.GetString(config.WalletUnlockPasswordFile)
+	runOnOnePort                  = config.GetBool(config.RunOnOnePortKey)
```
tiero, line 62: what if we infer this when trade and operator port are equal?

config/config.go

cmd/tdexd/main.go
sekulicd: Done

PR #625, opened 2022-08-25
Tdexconnect add proto & host to the url

### LINKED ISSUE #622: tdexdconnect: Support TLS termination with CA signed certificate
## Overview 

In some scenarios, the daemon could be deployed in a context where the TLS termination is already present (ie. from a reverse proxy on top) therefore it conflicts with the self-signed certificate of the operator interface.

We now can disable the generation of TLS self-signed via `TDEX_NO_OPERATOR_TLS` env var, but this makes generating tdexdconnect URLS like 

`tdexdconnect://<host>:<port>` if not initialized and `tdexdconnect://<host>:<port>?macaroon=` after seed has been init without the `cert` field.

We need to think of a way to communicate this to clients such as CLI or dashboard, in a way the client knows to use secure connection, but leave to the http client used to make the TLS handshake correctly.

## Proposal
One idea could be to introduce a new field in the query string such as `?proto=http` or `?proto=https` ?  

#### Validation/Behavior
- If the `?cert` has been given already we default to `https` and ignore anything given to the proto field ? 
- or we return error if `cert` is given, but anything beside `https` inside the `proto` field is passed?

@Janaka-Steph @sekulicd @louisinger

sekulicd:
This adds protocol and host to the tdex connect url.

This closes #622 

@tiero please review.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -97,6 +97,12 @@ const (
 	WalletUnlockPasswordFile = "WALLET_UNLOCK_PASSWORD_FILE"
 	// NoOperatorTlsKey is used to start the daemon without using TLS for the operator service
 	NoOperatorTlsKey = "NO_OPERATOR_TLS"
+	// ConnectUrlHost ectUrlHost is the host of the tdex-daemon service to connect to
+	//used by tdexconnect to create connection url
+	ConnectUrlHost = "CONNECT_URL_HOST"
+	// ConnectUrlProto is the http/https protocol of the tdex-daemon service to connect to
+	//used by tdexconnect to create connection url
+	ConnectUrlProto = "CONNECT_URL_PROTO"
 
 	DbLocation        = "db"
 	TLSLocation       = "tls"
@@ -105,6 +111,8 @@ const (
 
 	MinDefaultPercentageFee = 0.01
 	MaxDefaultPercentageFee = float64(99)
+
+	httpsProtocol = "https"
 )
 
 var vip *viper.Viper
@@ -137,6 +145,7 @@ func init() {
 	vip.SetDefault(RescanRangeStartKey, 0)
 	vip.SetDefault(RescanGapLimitKey, 50)
 	vip.SetDefault(NoOperatorTlsKey, false)
+	vip.SetDefault(ConnectUrlProto, httpsProtocol)
 
 	if err := validate(); err != nil {
 		log.Fatalf("error while validating config: %s", err)
--- a/internal/interfaces/http/tdex_connect_handler.go
+++ b/internal/interfaces/http/tdex_connect_handler.go
@@ -2,6 +2,8 @@ package httpinterface
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"html/template"
 	"io/ioutil"
 	"net/http"
@@ -18,8 +20,9 @@ import (
 )
 
 const (
-	templateFile = "web/layout.html"
-	cssFile      = "web/bulma.min.css"
+	templateFile  = "web/layout.html"
+	httpProtocol  = "http"
+	httpsProtocol = "https"
 )
 
 type TdexConnectService interface {
@@ -32,31 +35,42 @@ type tdexConnect struct {
 	walletUnlockerSvc application.WalletUnlockerService
 	macaroonBytes     []byte
 	certBytes         []byte
-	serverPort        string
 	macaroonPath      string
 	certPath          string
+	scheme            string
+	serverAddress     string
 }
 
 func NewTdexConnectService(
 	repoManager ports.RepoManager,
 	walletUnlockerSvc application.WalletUnlockerService,
-	macaroonPath, certPath, serverPort string,
-) TdexConnectService {
+	macaroonPath, certPath, serverPort, host, protocol string,
+) (TdexConnectService, error) {
 	macBytes := readFile(macaroonPath)
 	certBytes := readFile(certPath)
+
+	scheme := protocol
+	if certBytes != nil {
+		if protocol == "" || protocol == httpsProtocol {
+			scheme = httpsProtocol
+		} else if protocol == httpProtocol {
+			return nil, errors.New("http protocol invalid with cert provided")
+		}
+	}
 
 	return &tdexConnect{
 		walletUnlockerSvc: walletUnlockerSvc,
 		macaroonBytes:     macBytes,
 		certBytes:         certBytes,
-		serverPort:        serverPort,
 		macaroonPath:      macaroonPath,
 		certPath:          certPath,
 		repoManager:       repoManager,
-	}
+		scheme:            scheme,
+		serverAddress:     fmt.Sprintf("%s:%s", host, serverPort),
+	}, nil
 }
 
-func (h *tdexConnect) RootHandler(w http.ResponseWriter, req *http.Request) {
+func (t *tdexConnect) RootHandler(w http.ResponseWriter, req *http.Request) {
 	data := Page{
 		Title: "TDEX Daemon",
 	}
@@ -68,7 +82,6 @@ func (h *tdexConnect) RootHandler(w http.ResponseWriter, req *http.Request) {
 }
 
 func (t *tdexConnect) AuthHandler(w http.ResponseWriter, req *http.Request) {
-
 	ctx := context.Background()
 
 	cert := t.certBytes
@@ -79,7 +92,7 @@ func (t *tdexConnect) AuthHandler(w http.ResponseWriter, req *http.Request) {
 
 	// start building the TDEXD connect URL
 	connectUrl, err := tdexdconnect.EncodeToString(
-		t.serverPort, cert, nil,
+		t.scheme, t.serverAddress, cert, nil,
 	)
 	if err != nil {
 		log.Errorln(err.Error())
@@ -133,7 +146,7 @@ func (t *tdexConnect) AuthHandler(w http.ResponseWriter, req *http.Request) {
 
 	// append the macaroon
 	connectUrl, err = tdexdconnect.EncodeToString(
-		t.serverPort, cert, macaroon,
+		t.scheme, t.serverAddress, cert, macaroon,
 	)
 	if err != nil {
 		log.Errorln(err.Error())
--- a/pkg/tdexdconnect/tdexdconnect.go
+++ b/pkg/tdexdconnect/tdexdconnect.go
@@ -9,9 +9,10 @@ import (
 
 // Encode encodes the given args as query parameters of the returned *url.URL.
 func Encode(
-	rpcServerAddr string, certBytes, macBytes []byte,
+	schema, rpcServerAddr string,
+	certBytes, macBytes []byte,
 ) (*url.URL, error) {
-	u := url.URL{Scheme: "tdexdconnect", Host: rpcServerAddr}
+	u := url.URL{Scheme: schema, Host: rpcServerAddr}
 	q := u.Query()
 
 	if len(certBytes) > 0 {
@@ -34,9 +35,9 @@ func Encode(
 
 // EncodeToString encodes the given args into a base64 string URL.
 func EncodeToString(
-	rpcServerAddr string, certBytes, macBytes []byte,
+	schema, rpcServerAddr string, certBytes, macBytes []byte,
 ) (string, error) {
-	u, err := Encode(rpcServerAddr, certBytes, macBytes)
+	u, err := Encode(schema, rpcServerAddr, certBytes, macBytes)
 	if err != nil {
 		return "", err
 	}
@@ -46,7 +47,7 @@ func EncodeToString(
 // Decode decodes a base64 string URL and returns its query parameters.
 func Decode(
 	connectUrl string,
-) (rpcAddress string, certBytes, macBytes []byte, err error) {
+) (schema, rpcAddress string, certBytes, macBytes []byte, err error) {
 	u, err := url.Parse(connectUrl)
 	if err != nil {
 		return
@@ -66,6 +67,7 @@ func Decode(
 		return
 	}
 
+	schema = u.Scheme
 	rpcAddress = u.Host
 	certBytes = cBytes
 	macBytes = mBytes
--- a/cmd/tdex/config.go
+++ b/cmd/tdex/config.go
@@ -153,7 +153,7 @@ func configConnectAction(c *cli.Context) (err error) {
 		return
 	}
 
-	rpcServerAddr, certificate, macaroon, err :=
+	schema, rpcServerAddr, certificate, macaroon, err :=
 		tdexdconnect.Decode(connectUrl)
 	if err != nil {
 		return
@@ -198,6 +198,7 @@ func configConnectAction(c *cli.Context) (err error) {
 	noMacaroons := strconv.FormatBool(len(certificate) == 0 && len(macaroon) == 0)
 
 	if err = setState(map[string]string{
+		"schema":         schema,
 		"rpcserver":      rpcServerAddr,
 		"no_macaroons":   noMacaroons,
 		"tls_cert_path":  tlsCertPath,
--- a/cmd/tdexd/main.go
+++ b/cmd/tdexd/main.go
@@ -60,6 +60,8 @@ var (
 	rescanGapLimit                = config.GetInt(config.RescanGapLimitKey)
 	walletUnlockPasswordFile      = config.GetString(config.WalletUnlockPasswordFile)
 	noOperatorTls                 = config.GetBool(config.NoOperatorTlsKey)
+	host                          = config.GetString(config.ConnectUrlHost)
+	protocol                      = config.GetString(config.ConnectUrlProto)
 
 	version = "dev"
 	commit  = "none"
@@ -278,6 +280,8 @@ func NewGrpcService(
 			NoTls:                    noOperatorTls,
 			ExtraIPs:                 operatorTLSExtraIPs,
 			ExtraDomains:             operatorTLSExtraDomains,
+			Host:                     host,
+			Protocol:                 protocol,
 		}
 
 		return grpcinterface.NewServiceOnePort(opts)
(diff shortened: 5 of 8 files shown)


internal/interfaces/http/tdex_connect_handler.go
```diff
@@ -32,31 +35,42 @@ type tdexConnect struct {
...
 	certBytes := readFile(certPath)
 
+	scheme := protocol
+	if certBytes != nil {
+		if protocol == "" || protocol == httpsProtocol {
+			scheme = httpsProtocol
+		} else if protocol == httpProtocol {
+			return nil, errors.New("http protocol invalid with cert provided")
+		}
+	}
```
altafan, line 59: we should move this validation logic to the encoding function of `pkg/tdexdconnect` to avoid rewriting the same code where else that package is used and also to enforce that it generates only correct URLs.

tiero: Maybe we need also a Validate function of sort? so we stop before trying to encoding the string

sekulicd: Done

tiero (approved):

pkg/tdexdconnect/tdexdconnect.go
```diff
@@ -35,9 +36,13 @@ func Encode(
 
 // EncodeToString encodes the given args into a base64 string URL.
 func EncodeToString(
-	schema, rpcServerAddr string, certBytes, macBytes []byte,
+	scheme, rpcServerAddr string, certBytes, macBytes []byte,
 ) (string, error) {
-	u, err := Encode(schema, rpcServerAddr, certBytes, macBytes)
+	if certBytes != nil && scheme == "http" {
+		return "nil", errors.New("http protocol invalid with cert provided")
+	}
```
altafan, line 43: this should be in `Encode` and also `Decode`.

altafan (changes requested):
since the daemon's operator interface can now be served with macaroon but without tls, the `--insecure` flag of the tdexdconnect service must be replaced with 2 flags `--no-tls` and `--no-macaroon` to have proper control on what to include in the connect URL that is gonna be generated.

tiero (changes requested):
`pkg/tdexdconnect/tdexdconnect_test.go` must be updated

pkg/tdexdconnect/tdexdconnect.go
sekulicd: Done

pkg/tdexdconnect/tdexdconnect_test.go
```diff
@@ -34,31 +34,34 @@ Wwq8JtrendvmccXB
...
...edURL: "https://localhost:9000?cert=MIICqDCCAk6gAwIBAgIRAM4trYu7swKO7E9OI5q350YwCgYIKoZIzj0EAwIwQjENMAsGA1UEChMEdGRleDExMC8GA1UEAxMoTUJQZGlQaXJhbGJlcnRvLmhvbWVuZXQudGVsZWNvbWl0YWxpYS5pdDAeFw0yMTEwMDYxNTMzMDRaFw0yMjEwMDcxNTMzMDRaMEIxDTALBgNVBAoTBHRkZXgxMTAvBgNVBAMTKE1CUGRpUGlyYWxiZXJ0by5ob21lbmV0LnRlbGVjb21pdGFsaWEuaXQwWTATBgcqhkjOPQIBBggqhkjOPQMBBwNCAAR1WPEU-MKQVHNAdK0bG89FGsU6Sw6Izqe1taOfS0hBQfP-W2YBccLwooIKmPdjtZAoV4kZcguksW6gAdzgtQrmo4IBIzCCAR8wDgYDVR0PAQH_BAQDAgKkMA8GA1UdEwEB_wQFMAMBAf8wHQYDVR0OBBYEFI5lIKaTMgXPe0v8k9oRakHlPHtoMIHcBgNVHREEgdQwgdGCKE1CUGRpUGlyYWxiZXJ0by5ob21lbmV0LnRlbGVjb21pdGFsaWEuaXSCCWxvY2FsaG9zdIIEdW5peIIKdW5peHBhY2tldIcEfwAAAYcQAAAAAAAAAAAAAAAAAAAAAYcQ_oAAAAAAAAAAAAAAAAAAAYcQ_oAAAAAAAAAQCEC_uSHHLIcEwKgB14cQ_oAAAAAAAACcVyn__lDMuocQ_oAAAAAAAADp7kZyxh-R8IcQ_oAAAAAAAABBBISbVTjXoYcQ_oAAAAAAAACu3kj__gARIjAKBggqhkjOPQQDAgNIADBFAiEAv-StshVAd-iSAz_2oGC0e076aiVvWHgKSehisPugrngCIGQ3tjiqzC1-oxNNMFvr7OD4CAkbWwq8JtrendvmccXB&macaroon=AgEFdGRleGQChQEDChCqCBfWunS_Tu_-IsVdx6XxEgEwGhUKBm1hcmtldBIEcmVhZBIFd3JpdGUaFwoIb3BlcmF0b3ISBHJlYWQSBXdyaXRlGg4KBXByaWNlEgV3cml0ZRoVCgZ3YWxsZXQSBHJlYWQSBXdyaXRlGhYKB3dlYmhvb2sSBHJlYWQSBXdyaXRlAAAGINX1smi9KiBJiORAbg2aY5wNJ0dk45Uz4Iy3gIyGHgUC",
```
altafan, line 48: I thought we wanted to add a new param to the URL like `proto=http|https`, instead we are replacing the initial `tdexdconnect://` with the scheme protocol right away. Is this intended? @tiero

cmd/tdexdconnect/main.go
```diff
@@ -170,7 +176,7 @@ func main() {
 	}
 
 	connectUrl, err := tdexdconnect.EncodeToString(
-		rpcServerAddr, certBytes, macBytes,
+		schema, rpcServerAddr, certBytes, macBytes,
```
tiero, line 179: shoudn't here be used the ConnectUrlHost instead of rpcserver address?

config/config.go
```diff
@@ -97,6 +97,12 @@ const (
 	WalletUnlockPasswordFile = "WALLET_UNLOCK_PASSWORD_FILE"
 	// NoOperatorTlsKey is used to start the daemon without using TLS for the operator service
 	NoOperatorTlsKey = "NO_OPERATOR_TLS"
+	// ConnectUrlHost ectUrlHost is the host of the tdex-daemon service to connect to
+	//used by tdexconnect to create connection url
+	ConnectUrlHost = "CONNECT_URL_HOST"
+	// ConnectUrlProto is the http/https protocol of the tdex-daemon service to connect to
+	//used by tdexconnect to create connection url
+	ConnectUrlProto = "CONNECT_URL_PROTO"
```
tiero, line 105: we should set the default here === `127.0.0.1:` + OPERATOR_LISTENING_PORT

internal/interfaces/http/tdex_connect_handler.go
```diff
@@ -133,7 +146,7 @@ func (t *tdexConnect) AuthHandler(w http.ResponseWriter, req *http.Request) {
 
 	// append the macaroon
 	connectUrl, err = tdexdconnect.EncodeToString(
-		t.serverPort, cert, macaroon,
+		t.scheme, t.serverAddress, cert, macaroon,
```
tiero, line 149: shoudnt here user the config OPERATOR_URL_HOST?

tiero: If I pass then `TDEX_CONNECT_URL_HOST=127.0.0.1:9000` it will badly concatednate server addr with the host

image

pkg/tdexdconnect/tdexdconnect_test.go
tiero: Yes this wrong indeed. We need a new querystring param `proto`

`tdexdconnect` must reamain as scheme

sekulicd:
> since the daemon's operator interface can now be served with macaroon but without tls, the `--insecure` flag of the tdexdconnect service must be replaced with 2 flags `--no-tls` and `--no-macaroon` to have proper control on what to include in the connect URL that is gonna be generated.

Done

cmd/tdexdconnect/main.go
altafan: the tdexdconnect binary does not have access to daemon's configuration.. rather, the daemon should use it to build the correct `rpcServerAddr`.

internal/interfaces/http/tdex_connect_handler.go
altafan: in ref to [this comment](https://github.com/tdex-network/tdex-daemon/pull/625#discussion_r954812264), here we should override the host of server address in case `CONNECT_URL_HOST` is defined.

pkg/tdexdconnect/tdexdconnect_test.go
tiero: to be clear this shoud be a complete valid URLs

- `tdexdconnect://<host>?proto=https&cert=<cert>&macaroon=<macaroon>`
- `tdexdconnect://<host>?proto=https&macaroon=<macaroon>`
- `tdexdconnect://<host>?proto=http&macaroon=<macaroon>`
- `tdexdconnect://<host>?proto=http`

sekulicd: url.Sheme (pkg url) should to be used for protocol(ws, http, https)
@tiero why we use tdexdconnect ?

config/config.go
altafan: I do think it's ok to not have a default for `CONNECT_URL_HOST`. When the daemon builds the address for the connect URL, it default's to server one unless this env var has been set.

pkg/tdexdconnect/tdexdconnect_test.go
tiero: and maybe also 

- `tdexdconnect://<host>?proto=https`
- `tdexdconnect://<host>?proto=http&cert=<cert>`

but not sure how valuable it is.

config/config.go
tiero: ok, it will be something like `:9000` then, since we do not know the hostname

internal/interfaces/http/tdex_connect_handler.go
tiero: ok then we must pass `CONNECT_URL_HOST` to this method

cmd/tdexdconnect/main.go
tiero: Yes what I mean is inside tdexd where we generate the HTML page to `rpcServerAddr` we must pass the connect URL, not the OPERATOR_LISTENING_PORT

pkg/tdexdconnect/tdexdconnect_test.go
tiero: > @tiero why we use tdexdconnect ?

This is our scheme we created, to make it clear is tdexdconnect URL. ALways has been

tiero: example: https://github.com/LN-Zap/lndconnect/blob/master/lnd_connect_uri.md

This makes sure it's used by "TDEXD clients" and not normal http clients such as browsers, curls etc...

tiero: you can use any scheme you want with pkg url

https://github.com/LN-Zap/lndconnect/blob/master/lndconnect.go#L71

config/config.go
tiero: If clients see a host like `:9000` it's kind of standard to assume is localhost, just saying we can eventually do this a tdexdconnect pkg level?

sekulicd:
@tiero @altafan please review again.

internal/interfaces/http/tdex_connect_handler.go
```diff
@@ -32,31 +35,47 @@ type tdexConnect struct {
...
+			p = httpsProtocol
+		} else if protocol == httpProtocol {
+			return nil, errors.New("http protocol invalid with cert provided")
+		}
+	}
+
+	h := host
+	if host == "" {
+		h = "localhost"
+	}
```
altafan, line 64: `CONNECT_URL_HOST` should be used in `AuthHandler` to build the correct address instead of always using `t.serverAddress`. I think it's ok for this service to have a simple `serverPort` field instead `serverAddress`.

altafan: ref to [this comment](https://github.com/tdex-network/tdex-daemon/pull/625#discussion_r954813407).

altafan:
@sekulicd @tiero I think in this PR we're forgetting about the case where the Operator service has TLS enabled with self-signed cert and extra_ip/extra_domain set.
As already discussed with dusan, the simplest solution is to override `host` in `cmd/tdexd/main.go` in case an operator extra domain or IP is set (by giving precedence to domain since both can be set).
At the same time, in config we should enforce that connectUrl host and operator extra ip/domain are mutually exclusive since the first means using a CA cert while the second ones mean using a self-signed cert for the operator interface.

sekulicd:
> @sekulicd @tiero I think in this PR we're forgetting about the case where the Operator service has TLS enabled with self-signed cert and extra_ip/extra_domain set. As already discussed with dusan, the simplest solution is to override `host` in `cmd/tdexd/main.go` in case an operator extra domain or IP is set (by giving precedence to domain since both can be set). At the same time, in config we should enforce that connectUrl host and operator extra ip/domain are mutually exclusive since the first means using a CA cert while the second ones mean using a self-signed cert for the operator interface.

@tiero @altafan please check if this is ok.

tiero:
> At the same time, in config we should enforce that connectUrl host and operator extra ip/domain are mutually exclusive since the first means using a CA cert while the second ones mean using a self-signed cert for the operator interface.

Don't think so, they are not mutually exclusive, but it's correct, if both given, they can't be different otherwise won't work.

What I would do is instead return error when given if they are different, in way that at least either IP or DOMAIN must be euqal to HOST.

 extra_ip/extra_domain should only affect behvior of generation of self signed cert, and CONNECT_URL_HOST should only affect what is shown in the HTML Page.

config/config.go
```diff
@@ -97,6 +97,12 @@ const (
 	WalletUnlockPasswordFile = "WALLET_UNLOCK_PASSWORD_FILE"
 	// NoOperatorTlsKey is used to start the daemon without using TLS for the operator service
 	NoOperatorTlsKey = "NO_OPERATOR_TLS"
+	// ConnectUrlHost ectUrlHost is the host of the tdex-daemon service to connect to
```
tiero, line 100:
```suggestion
	// ConnectUrlHost is the host of the tdex-daemon service to connect to ie. myservice:9000
```

config/config.go
```diff
@@ -318,9 +327,35 @@ func validate() error {
 		}
 	}
 
+	if vip.GetString(OperatorExtraDomainKey) != "" ||
+		vip.GetString(OperatorExtraIPKey) != "" {
+		if vip.GetString(ConnectUrlHost) != "" {
+			return fmt.Errorf("%s and %s/%s cannot be set together",
```
tiero, line 333: imho they CAN be set together, we should check the hostname to be either OperatorExtraDomainKey or OperatorExtraIPKey.

Beaware that host != hostname, we should extract and left out the port. (AFAIK std library helps on this)

config/config.go
```diff
@@ -318,9 +327,35 @@ func validate() error {
...
+	if vip.GetString(OperatorExtraDomainKey) != "" {
+		host = vip.GetString(OperatorExtraDomainKey)
+	}
+
+	if host == "" {
+		host = "localhost"
+	}
+
+	return host
+}
```
tiero, line 357: assuming we pass [this](https://github.com/tdex-network/tdex-daemon/pull/625/files#r955216201) check we should give priority to ConnectUrlHost, if not given domain should have prioirity over ip (if both given).

internal/interfaces/http/tdex_connect_handler.go
```diff
@@ -32,31 +35,42 @@ type tdexConnect struct {
...
+) (TdexConnectService, error) {
 	macBytes := readFile(macaroonPath)
 	certBytes := readFile(certPath)
 
+	p := protocol
+	if len(certBytes) > 0 {
+		if protocol == "" || protocol == httpsProtocol {
+			p = httpsProtocol
+		} else if protocol == httpProtocol {
+			return nil, errors.New("http protocol invalid with cert provided")
```
tiero, line 57:
```suggestion
			return nil, errors.New("tdexdconnect: proto=http is not valid if cert is given")
```

config/config.go

config/config.go

config/config.go

internal/interfaces/http/tdex_connect_handler.go
sekulicd: Done

tiero (approved):

config/config.go
```diff
@@ -321,6 +330,54 @@ func validate() error {
 	return nil
 }
 
+func GetHostname() (string, error) {
+	return getHostname(
+		vip.GetString(OperatorExtraIPKey),
+		vip.GetString(OperatorExtraDomainKey),
```
altafan, line 336: These vars are actually string slices, not string
```suggestion
		vip.GetStringSlice(OperatorExtraIPKey),
		vip.GetStringSlice(OperatorExtraDomainKey),
```

config/config.go
```diff
@@ -321,6 +330,54 @@ func validate() error {
...
+		vip.GetString(ConnectUrlHost),
+	)
+}
+
+func getHostname(
+	operatorExtraIP string,
+	operatorExtraDomain string,
```
altafan, line 343:
```suggestion
	operatorExtraIP, operatorExtraDomain []string,
```

config/config.go
```diff
@@ -321,6 +330,54 @@ func validate() error {
...
+		hostname = ""
+	)
+
+	if operatorExtraIP != "" {
+		host = operatorExtraIP
+	}
+
+	if operatorExtraDomain != "" {
+		host = operatorExtraDomain
+	}
```
altafan, line 357:
```suggestion
	if len(operatorExtraIP) > 0 {
		host = operatorExtraIP[0]
	}

	if len(operatorExtraDomain) > 0 {
		host = operatorExtraDomain[0]
	}
```

config/config.go

config/config.go

config/config.go

PR #629, opened 2022-09-12
Rename env vars & Fix tdexconnect url

### LINKED ISSUE #627: tdexdconnect html page: wrong port and URL format
Running a daemon from master will have you have an URL with TDEX_CONNECT_URL_HOST=localhost:3000 will have you ignore that and returns something like `tdexdconnect://localhost::9000` ignoring the passed host and concatenaing `:` two times

altafan:
This renames the variables for tdexconnect URL to `TDEX_CONNECT_ADDR` and `TDEX_CONNECT_PROTO`. The first defaults to empty while the second to https given the default configuration with operator tls enabled.

If an operator extra tls IP or domain is already there's no need to define a connect address unless the service is not reachable on the default port.

By fixing the logic of the server address to connect to, this closes #627.

Please @tiero review this.


cmd/tdexd/main.go
```diff
@@ -262,9 +263,13 @@ func NewGrpcService(
...
 	repoManager ports.RepoManager,
 ) (interfaces.Service, error) {
-	hostname, err := config.GetHostname()
-	if err != nil {
-		return nil, err
+	addr := fmt.Sprintf("localhost:%d", operatorSvcPort)
+	if len(operatorTLSExtraDomains) > 0 {
+		addr = operatorTLSExtraDomains[0]
+	} else if len(operatorTLSExtraIPs) > 0 {
+		addr = operatorTLSExtraIPs[0]
```
tiero, line 272: operatorTLSExtraDomains and operatorTLSExtraIPs should only be used for self-signed cert generation imho

config/config.go
```diff
@@ -98,11 +97,11 @@ const (
 	WalletUnlockPasswordFile = "WALLET_UNLOCK_PASSWORD_FILE"
 	// NoOperatorTlsKey is used to start the daemon without using TLS for the operator service
 	NoOperatorTlsKey = "NO_OPERATOR_TLS"
-	// ConnectUrlHost is the host of the tdex-daemon service to connect to ie. myservice:9000
-	ConnectUrlHost = "CONNECT_URL_HOST"
-	// ConnectUrlProto is the http/https protocol of the tdex-daemon service to connect to
-	//used by tdexconnect to create connection url
-	ConnectUrlProto = "CONNECT_URL_PROTO"
+	// ConnectAddrKey is the address <host:port> of the tdex-daemon service to connect to, ie. myservice:9000
```
tiero, line 100: there is a semantic problem I guess: address should include protocol (http,https) meanwhile host is just hostname:address?

altafan: > [...] meanwhile host is just hostname:address?

i'm sure you were meant hostname:port.. beside this, the naming I use comes from pkg [net/http](https://pkg.go.dev/net/http#Server).. I would not change it just for consistency.

PR #636, opened 2022-09-22
401 response should contain CORS and "WWW-Authenticate" headers

Janaka-Steph:
StatusUnauthorized 401 error response should always have a specific "WWW-Authenticate" header https://developer.mozilla.org/fr/docs/Web/HTTP/Headers/WWW-Authenticate
Also should reply with CORS headers.

This allow the frontend running in different origin (port 3003) to get the status error code and not a generic error like `Failed to fetch`. See: https://github.com/axios/axios/issues/960

Please review @tiero @altafan


altafan:
@Janaka-Steph please, add this line to Encrypt/Decrypt functions.

Also, it would be enough to check that the hash of the given password matches the `vault`'s one instead of actually decrypting the encrypted mnemonic.

PR #644, opened 2022-09-30
Fix GetMarketReport predefined period all

### LINKED ISSUE #605: PREDEFINED_PERIOD_ALL is stoping at previous calendar year
All 2022 data points are missing.

sekulicd:
This fixes bug that was happening when GetMarketReport RPC was invoked with predefined period PREDEFINED_PERIOD_ALL.

This closes #605 

@tiero @altafan @Janaka-Steph please review.

### DIFF
--- a/internal/core/application/types.go
+++ b/internal/core/application/types.go
@@ -28,8 +28,8 @@ const (
 	LastMonth
 	LastThreeMonths
 	YearToDate
-	All
 	LastYear
+	All
 
 	StartYear = 2021
 )
--- a/internal/core/application/types.go
+++ b/internal/core/application/types.go
@@ -543,7 +543,7 @@ type PredefinedPeriod int
 
 func (p *PredefinedPeriod) validate() error {
 	if *p > All {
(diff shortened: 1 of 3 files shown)


altafan:
@sekulicd please add a test that demonstrates the bug is fixed.

sekulicd:
> @sekulicd can you please add a test that demonstrates the bug is fixed?

Added test for application layer GetMarketReport

internal/core/application/types.go
```diff
@@ -542,7 +542,7 @@ func (t *TimeRange) Validate() error {
 type PredefinedPeriod int
 
 func (p *PredefinedPeriod) validate() error {
-	if *p > LastYear {
+	if *p > All {
 		return fmt.Errorf("PredefinedPeriod cant be > %v", LastYear)
```
altafan, line 546:
```suggestion
		return fmt.Errorf("PredefinedPeriod cant be > %v", All)
```

sekulicd: Done

tiero (approved):

PR #648, opened 2022-10-07
TLS unset isCA indicator

### LINKED ISSUE #645: Invalid self-signed certificate for operator interface
We must unset [IsCA](https://github.com/tdex-network/tdex-daemon/blob/master/internal/interfaces/grpc/http_utils.go#L149) when generating the SS certificate for the operator server.

We might want to explore also if it's possible to remove its expiration date. Currently, it is set to [1 year](https://github.com/tdex-network/tdex-daemon/blob/master/internal/interfaces/grpc/http_utils.go#L70)

sekulicd:
This unsets isCA certificate indicator.

Regarding: "We might want to explore also if it's possible to remove its expiration date. Currently, it is set to [1 year](https://github.com/tdex-network/tdex-daemon/blob/master/internal/interfaces/grpc/http_utils.go#L70)" -> I dont see any validation preventing this.

This closes #645 

### DIFF
--- a/internal/interfaces/grpc/http_utils.go
+++ b/internal/interfaces/grpc/http_utils.go
@@ -146,7 +146,6 @@ func generateOperatorTLSKeyCert(
 
 		KeyUsage: x509.KeyUsageKeyEncipherment |
 			x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign,
-		IsCA:                  true,
 		BasicConstraintsValid: true,
 
 		DNSNames:    dnsNames,


altafan:
@sekulicd if it's not a problem, let's remove the expiration date from the cert then.

sekulicd:
> @sekulicd if it's not a problem, let's remove the expiration date from the cert then.

Sorry, I made a mistake, we cant remove it since if removed default time(0001-01-01T00:00:00Z) is set for both before/after and there is verification in [here](https://github.com/golang/go/blob/747e1961e95c2eb3df62e045b90b111c2ceea337/src/crypto/x509/verify.go#L574).
So we cant remove expiration time, verified by testing as well.

altafan (approved):
tACK 26c69e47a31216f88320d31535df653fbd41d215

PR #649, opened 2022-10-07
tdexconnect support proto=http

### LINKED ISSUE #643: tdex cli: config connect does not support proto=http
I tried to pass `tdexdconnect://localhost:9000?proto=http` as connect url to `tdex config connect` command, but 
then calling `tdex status` still returns an error that is looking for a TLS certificate

```sh
tdex status
[tdex] could not read TLS certificate:  open : no such file or directory
exit status 1
```

## Workaround

`tdex config set no_tls true`

sekulicd:
This sets no_tls flag to true if proto=http.

This closes #643 

### DIFF
--- a/cmd/tdex/main.go
+++ b/cmd/tdex/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"crypto/tls"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -289,15 +290,21 @@ func getClientConn(skipMacaroon bool) (*grpc.ClientConn, error) {
 			)
 		}
 
-		tlsCreds := credentials.NewTLS(nil)
+		config := &tls.Config{
+			InsecureSkipVerify: true,
+			MinVersion:         tls.VersionTLS12,
+		}
+		dialOpt := grpc.WithTransportCredentials(credentials.NewTLS(config))
 		if certPath != "" {
-			tlsCreds, err = credentials.NewClientTLSFromFile(certPath, "")
+			tlsCreds, err := credentials.NewClientTLSFromFile(certPath, "")
 			if err != nil {
 				return nil, fmt.Errorf("could not read TLS certificate:  %s", err)
 			}
+
+			dialOpt = grpc.WithTransportCredentials(tlsCreds)
 		}
 
-		opts = append(opts, grpc.WithTransportCredentials(tlsCreds))
+		opts = append(opts, dialOpt)
 	}
 
 	noMacaroons, _ := strconv.ParseBool(state["no_macaroons"])
--- a/cmd/tdex/main.go
+++ b/cmd/tdex/main.go
@@ -291,7 +291,7 @@ func getClientConn(skipMacaroon bool) (*grpc.ClientConn, error) {
 		}
 
 		config := &tls.Config{
-			InsecureSkipVerify: true,
+			InsecureSkipVerify: true, //nolint:gosec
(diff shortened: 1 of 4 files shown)


altafan:
@sekulicd this is not enough to close the ticket.

We need also to handle [these options](https://github.com/tdex-network/tdex-daemon/issues/643#issuecomment-1263742086) when [creating the connection](https://github.com/tdex-network/tdex-daemon/blob/2f51e824825b1730901606bc93a7b87d1277ba20/cmd/tdex/main.go#L280-L297). At the moment we support 1) and 3) but not 2)

sekulicd:
> @sekulicd this is not enough to close the ticket.
> 
> We need also to handle [these options](https://github.com/tdex-network/tdex-daemon/issues/643#issuecomment-1263742086) when [creating the connection](https://github.com/tdex-network/tdex-daemon/blob/2f51e824825b1730901606bc93a7b87d1277ba20/cmd/tdex/main.go#L280-L297). At the moment we support 1) and 3) but not 2)

Is it ok now?

pkg/tdexdconnect/tdexdconnect.go
```diff
@@ -17,6 +17,10 @@ func Encode(
 		return nil, errors.New("http protocol invalid with cert provided")
 	}
 
+	if len(certBytes) == 0 && proto == "https" {
+		return nil, errors.New("https protocol invalid without cert provided")
+	}
+
```
altafan, line 23: Why? This should be allowed indeed. In case proto is https and there's no cert, then CA cert (`credentials.NewTls(nil)`) should be used.

sekulicd: Then [this](https://github.com/sekulicd/tdex-daemon/blob/967d788e7f341319a58a65a9941cdd577f745d40/internal/interfaces/http/tdex_connect_handler.go#L66) is invalid as well no?

altafan: yes. In general, I thought e10bb6b49280d48b71198c23d2a19a3002b9249e was ok. What are later changes for?

sekulicd: we cant use credentials.NewTls(nil) for self signed cert since "...not trusted authority..." error is thrown.

sekulicd: @altafan checks removed.

cmd/tdex/main.go
```diff
@@ -289,11 +290,18 @@
 			)
 		}
 
-		tlsCreds, err := credentials.NewClientTLSFromFile(certPath, "")
-		if err != nil {
-			return nil, fmt.Errorf("could not read TLS certificate:  %s", err)
+		config := &tls.Config{InsecureSkipVerify: true, MinVersion: tls.VersionTLS12} //#nosec
```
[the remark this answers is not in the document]
tiero: add //nogosec comment to prevent this warning

sekulicd:
@tiero @altafan i reverted the change but when try to test it with stage env, i am getting:
[tdex] rpc error: code = Unimplemented desc = unexpected HTTP status code received from server: 404 (Not Found); transport: received unexpected content-type “text/plain; charset=utf-8”

tiero:
the error is due to the configuration of the reverse proxy where that instance is deployed. Is not introduced in this PR code either, already happening in master, so I would go ahead and merge this.

tiero (approved):

PR #650, opened 2022-10-07
add pathname to increase supports of all deployments

tiero:
If someone is exposing the daemon with some reverse proxy on ie. `/api` the web page would not have worked calling the `/tdexdconnect` endpoint on the main host not honoring base pathname

### DIFF
--- a/web/layout.html
+++ b/web/layout.html
@@ -33,8 +33,9 @@
         headers.set('Authorization', 'Basic ' + window.btoa(unescape(encodeURIComponent(authString))))
       }
 
-      const response = await window.fetch("/tdexdconnect", { method: 'GET', headers,});
+      const response = await window.fetch(`${window.location.pathname}tdexdconnect`, { method: 'GET', headers,});
       if (!response.ok) {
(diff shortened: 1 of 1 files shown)


web/layout.html
```diff
@@ -33,8 +33,9 @@
         headers.set('Authorization', 'Basic ' + window.btoa(unescape(encodeURIComponent(authString))))
       }
 
-      const response = await window.fetch("/tdexdconnect", { method: 'GET', headers,});
+      const response = await window.fetch(`${window.location.pathname}tdexdconnect`, { method: 'GET', headers,});
```
altafan, line 36: Not sure this is always correct. I tried out `window.location.pathname` and when it's not `/`, it retuns a string like `/path/example`. I think the best thing to do is making sure it's value ends with `/`:
```suggestion
      const pathname = window.location.pathname.endsWith('/') ? window.location.pathname : `${window.location.pathname}/`
      const response = await window.fetch(`${window.location.pathname}tdexdconnect`, { method: 'GET', headers,});
```

PR #669, opened 2023-01-16
Add GetMarketPrice API

altafan:
This adds a new rpc `GetMarketPrice` accordingly to the updated tdex-protobuf protos.

Please @tiero, review this.


altafan:
@tiero this adds also support for base/quote asset precision, required to return the correct min tradable amount eventually.

PR #685, opened 2023-03-20
Move to tdex-protobuf@v2 and tdex-daemon@v2

### LINKED ISSUE #670: [v1] Update tdex-protobuf to v2
We should let the trader pick which asset he wants to be charged the fee in a chosen market. (now we always charge in the asset "on the way in")

One caveat is that there should be a "fee policy" in place that allow operators to disable the fee to be charged in a specific asset, especially important in the case of securities where you MUST charge always in the quote asset (or in general in the non-security asset).

# Breaking change

This should bump the protocol to the V2, since the behavior of clients will then need to change along with the Swap protocol messages, where `amount_p` and `amount_r` will need to be the actual amount WITHOUT fee added/subtracted (depending on the tradeType) and the `fee_asset` and `fee_amount` should be explicit in the SwapRequest

altafan:
This contains changes to make the daemon (v1) supporting ONLY the new tdex-protobuf and tdex-daemon v2 protos, where:
- from trader POV, he can now select the asset in which paying the trading fees applied by a provider. The provider returns both a preview amount and a fee amount. These amounts, along with the one of the request message, should be used in the swap request message, which now contains 2 new fields `string fee_asset` and `uint64 fee_amount`. It's also up to the trader to correctly charge the fee to either the input or output amount when crafting the swap request tx.
- from operator POV, he can now set percentage fees on both assets on the market just like the fixed one (we'll add fee policies in future PR). Given this change, the env var for the default percentage fee has been dropped.
He can also set an optional name for a market. In the future, he will be able to refer to a market with this name, as an alternative to using both assets.

Closes #670.

Please @sekulicd @tiero review this.


tiero:
cc/ @Janaka-Steph

pkg/swap/complete.go
```diff
@@ -18,45 +19,107 @@ type CompleteOpts struct {
...
+
+		if ptx, _ := psetv2.NewPsetFromBase64(opts.Transaction); ptx != nil {
+			ok, err := ptx.ValidateAllSignatures()
+			if err != nil {
+				return "", nil, err
+			}
+			if !ok {
+				return "", nil, fmt.Errorf("transaction contains invalid signatures")
+			}
+		} // TODO: validate sigs of raw tx input
```
Janaka-Steph, line 113: This comment can be removed?

altafan: nice catch, done.

internal/core/domain/market.go
```diff
@@ -422,37 +433,29 @@ func (m *Market) priceFromBalances(
...
-		return safeSubtractFeesFromAmount(
-			amount, uint64(m.FixedFee.QuoteFee), quoteBalance,
-		)
-	}
-	return amount + uint64(m.FixedFee.BaseFee), nil
-}
+	fee := decimal.NewFromInt(int64(percentageFee)).Div(decimal.NewFromInt(10000))
+	feeAmount := decimal.NewFromInt(int64(amount)).Mul(fee).BigInt().Uint64()
+	feeAmount += fixedFee
+	// Fees must always be added on the amount sent by the proposer and
```
sekulicd, line 447: this is very important comment for both devs and daemon clients(traders), as it contains business logic on fees, and it should maybe go to PreviewTrade/trade.proto doc.

internal/core/domain/market.go
```diff
@@ -11,9 +12,9 @@ import (
 	"github.com/tdex-network/tdex-daemon/pkg/mathutil"
 )
 
-type FixedFee struct {
-	BaseFee  uint64
-	QuoteFee uint64
+type MarketFee struct {
+	BaseAsset  uint64
```
sekulicd, line 16: Should we rename BaseAsset to BaseAssetFee? same for QuoteAsset

internal/core/domain/market.go
```diff
@@ -60,10 +63,10 @@ type Market struct {
 	BaseAssetPrecision uint
 	// Precison of the quote asset.
 	QuoteAssetPrecision uint
-	// Percentage fee expressed in basis points.
-	PercentageFee uint32
-	// Fixed fee amount expressed in satoshi for both assets.
-	FixedFee FixedFee
+	// Percentage fee expressed in basis points for both assets.
+	PercentageFee MarketFee
```
sekulicd, line 67: Should we use same struct(MarketFee) for both PercentageFee(expressed in basis points) and FixedFee(expressed in sats)?

examples/buy-lbtc-with-usdt/main.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+	client = tdexv2.NewTradeServiceClient(conn)
+
+	explorerSvc, err = esplora.NewService("http://localhost:3001", 15000)
+	if err != nil {
+		return fmt.Errorf("failed to prepare explorer: %s", err)
+	}
+	return nil
+}
+
```
sekulicd, line 47: Should we add doc that in order to run example it is needed to start daemon and setup market? or maybe make cmd(invoking script) that starts daemon with neccessery data?

internal/core/domain/market.go
altafan: it seems to me like a repetition based on the name given to the Market fields:
i think `market.PercentageFee.BaseAsset` is better than `market.PercentageFee.BaseAssetFee`.

internal/core/domain/market.go
altafan: You're definitely right. We must add comments on tdex-protobuf for sure and we must also update the official documentation and specs

internal/core/domain/market.go
altafan: Yes, that's what I did.

examples/buy-lbtc-with-usdt/main.go
altafan: My idea is to use these examples in the e2e test, which is yet to be fixed. Just wanted to do it in a separate PR.

Also, I didn't add a README for them because we would need to describe all the setup to open a market etc., while we already have the [dev-portal](https://dev.tdex.network/docs/provider/) for this, which, is yet to be updated too.

PR #688, opened 2023-03-23
[Draft]Local price feeder

sekulicd:
This is POC for incorporating price feeder inside daemon.
Its purpose is to discuss potential solutions.
One thing not yet clear is how do we bind tickers to base/asset pairs.
Do we update existing CreateMarket RPC and add ticker per asset? Do we add new RPC RegisterAssetTicker? or smtn else?

### DIFF
--- a/internal/core/application/operator/types.go
+++ b/internal/core/application/operator/types.go
@@ -65,6 +65,11 @@ type marketInfo struct {
 	balance map[string]ports.Balance
 }
 
+func (i marketInfo) Ticker() string {
+	//TODO implement me
+	panic("implement me")
+}
+
 func (i marketInfo) GetBaseAsset() string {
 	return i.BaseAsset
 }
@@ -133,6 +138,11 @@ func (s tradeStatus) IsFailed() bool {
 
 type tradeInfo struct {
 	domain.Trade
+}
+
+func (i tradeInfo) Ticker() string {
+	//TODO implement me
+	panic("implement me")
 }
 
 func (i tradeInfo) GetId() string {
@@ -404,6 +414,11 @@ type marketReportInfo struct {
 	marketFeeReportInfo
 	totVolume  marketVolumeInfo
 	subVolumes marketVolumeInfoList
+}
+
+func (i marketReportInfo) Ticker() string {
+	//TODO implement me
+	panic("implement me")
 }
 
 func (i marketReportInfo) GetMarket() ports.Market {
--- a/internal/core/application/price-updater/service.go
+++ b/internal/core/application/price-updater/service.go
@@ -1,0 +1,103 @@
+package priceupdater
+
+import (
+	"context"
+
+	"github.com/tdex-network/tdex-daemon/internal/core/domain"
+
+	log "github.com/sirupsen/logrus"
+
+	"github.com/tdex-network/tdex-daemon/internal/core/ports"
+)
+
+// Service periodically updates markets prices.
+// prices are fed into service from external price provider eg. Kraken.
+type Service struct {
+	// priceFeeder is the external price provider.
+	priceFeeder ports.PriceFeeder
+	// repoManager is the used to access the db.
+	repoManager ports.RepoManager
+}
+
+func NewService(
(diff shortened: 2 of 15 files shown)


altafan:
> Do we update existing CreateMarket RPC and add ticker per asset? Do we add new RPC RegisterAssetTicker? or smtn else?

I still need to check the code, but in general the ticker depends strictly on the price data source, like for example the market lbtc/usdt might have a different ticker on kraken rather than bitfinex or coinbase.

We could give the daemon's feeder component access to the domain repos to read data, but pheraps it needs a dedicated db where it can store info such as asset tickers, price sources, etc.

altafan:
@sekulicd I updated the target branch for this PR. You likely forgot to set it to `v1`, was pointing to `master`

altafan:
One of the key points of merging the feeder into the daemon is that we should be able to configure and interact with it via RPC calls. This means dropping the config file used by the original external service.

These could be the RPCs to be consumed by the operator:
```
service FeederService {
   rpc AddPriceFeed(AddPriceFeedRequest) returns (AddPriceFeedResponse);
   rpc StartPriceFeed(StartPriceFeedRequest) return (StartPriceFeedResponse);
   rpc StopPriceFeed(StartPriceFeedRequest) return (StopPriceFeedResponse);
   rpc UpdatePriceFeed(UpdatePriceFeedRequest) returns (UpdatePriceFeedResponse);
   rpc RemovePriceFeed(RemovePriceFeedRequest) returns (RemovePriceFeedResponse);
   rpc ListPriceFeeds(ListPriceFeedsRequest) returns (ListPriceFeedsResponse);
   rpc ListSupportedPriceSources(ListSupportedPriceSourcesRequest)
      returns (ListSupportedPriceSourcesResponse);
}

message AddPriceFeedRequest {
   Market market = 1;
   string source = 2;
   string ticker = 3;
}
message AddPriceFeedResponse {
   string id = 1;
}

message StartPriceFeedRequest {
   string id = 1;
}
message StartPriceFeedResponse {}

message StopPriceFeedRequest {
   string id = 1;
}
message StopPriceFeedResponse {}

message UpdatePriceFeedRequest {
   string id = 1;
   string source = 2;
   string ticker = 3;
}
message UpdatePriceFeedResponse {}


message RemovePriceFeedRequest {
   string id = 1;
}
message RemovePriceFeedResponse {}


message ListPriceFeedsRequest {
   Market market = 1;
}
message ListPriceFeedsResponse {
   repeated PriceFeed price_feeds = 1;
}

message ListSupportedPriceSourcesRequest {}
message ListSupportedPriceSourcesResponse {
   repeated string price_sources = 1;
}

message PriceFeed {
   string id = 1;
   Market market = 2;
   string source = 3;
   string ticker = 4;
}
```

With these RPCs, it would be up to the operator to specify the price source and the asset ticker for a market price feed and he could check the list of available sources with `ListSupportedPriceSources` at anytime.
These APIs give also a lot of flexibility to the operator because he could tweak the configuration of a market price feed at will (by always stopping it first, similar to the management of a market), for example to change the price source and ticker, or also to just change the ticker if he set a wrong one by mistake. Last but not least, a price feed could be removed if for example the operator changes strategy for the market and doesn't need the price feed anymore.

One (cool?) thing that just came up in my mind while writing would be detecting the change of strategy and automatically starting or stopping a price feed, if any exists for that specific market. Could be a feature to add for future iterations, just brainstorming ideas.

tiero:
Yes agreed on having RPCs to configure the feeder. I would retail the "well knonw pairs" 

Ie. We can hardcode that LBTC and USDT asset hash to known ticker for each API, so when you "attach" a feed to a market via RPC it will read the market and auronativally bind

sekulicd:
> Yes agreed on having RPCs to configure the feeder. I would retail the "well knonw pairs"
> 
> Ie. We can hardcode that LBTC and USDT asset hash to known ticker for each API, so when you "attach" a feed to a market via RPC it will read the market and auronativally bind

@tiero can you elaborate more on reasoning behind "well know pairs"? why do we want to kind override user input when he adds new feed to market?

Issue #657, opened 2022-12-12
[v1] feeder: move tdex-feeder service inside daemon, add RPCs to attach feed to market


tiero:

altafan:
Since now the daemon doensn't take care of managing the wallet anymore, the idea is that we can extend its scope to market-related operations like those performed currently by the feeder standalone service, ie. forwarding market price feeds taken from a source like for example kraken or bitfinex to the markets opened on the daemon that have PLUGGABLE strategy set.

This means that we can add new RPCs to the tdex-daemon protos  in order to replicate the original work done by the feeder and get rid of that service at the same time - we can prevent creating the price macaroon file made, in origin, ad hoc for it.

tiero:
I would still retain the price.macaroon and the external feeder, as advancement deployments can benefit to reuse custom feeder/externa software to push prices.

This only helps solo-operator and "one-shot" deployments, that will only have one single instance anyway

altafan:
> we should still have well known asset pairs and automatically bind to symbols (ie. XBT-USDT > L-BTC-USDT)

Not sure the well-known-market thing is a great feature actually. With all the stablecoins that are constantly being minted out there we must then keep this list of well-known assets up-to-date, which sounds annoying from a maintainability POV.

I would just keep it simple and since the operator would still need to provide both source and ticker for a non-well-known market, i think it's better to maintain this also for the well-known ones.

altafan:
Ref for api-spec and implementation: https://github.com/tdex-network/tdex-daemon/pull/688#issuecomment-1481417016

Logic for managing price feeds, inspired by the management of a market:
- A price feed can always be added to a market regardless its strategy, the only constraint is that there can't be more than one price feed per market.
- Starting a price feed requires the market to have PLUGGABLE strategy.
- To update a price feed it is mandatory to stop it first
- A price feed can be removed only if it's stopped and the strategy of the market is not PLUGGABLE

No more constraints for now since in a second iteration we will detect market strategy changes via events and make the feeder to properly stop a price feed for a market that has no more pluggable strategy.

altafan:
Closed by #691.

PR #691, opened 2023-03-29
Local price feeder 

sekulicd:
This incorporates price feeder inside daemon with the addition of new proto service that is in charge for configuring price feeds for markets.

This is in WIP state cause of few reasons:
1. Its not tested 
2. Swap pkg cant be compiled cause of wrong protobuf version, i need advise on which commit and need to cherry-pick or which new protobuf version to use @altafan 


altafan:
2) the original implementation of the price feeder service is designed to be configured only with a static json file, therefore there's no way to dynamically add or remove feeds for markets.
Since we are refactoring how to interact with it, let's just make changes to the source implementations to fit our new needs.

3) pkg/swap is fixed in #685. Once it gets merged you can sync with upstream/v1 and get rid of this problem.

api-spec/protobuf/tdex-daemon/v2/feeder.proto
```diff
@@ -0,0 +1,84 @@
+syntax = "proto3";
+
+package tdex_daemon.v2;
+
+import "tdex/v1/types.proto";
+import "tdex-daemon/v2/types.proto";
+
+// FeederService is the service to manage price feeds
+service FeederService {
```
altafan, line 9: currently missing `ListPriceFeeds` rpc to let the operator list the price feeds of all markets.

api-spec/protobuf/tdex-daemon/v2/feeder.proto
```diff
@@ -0,0 +1,84 @@
...
+  // AddPriceFeed creates a new price feed for the given market
+  rpc AddPriceFeed(AddPriceFeedRequest) returns (AddPriceFeedResponse);
+  // StartPriceFeed starts the price feed with the given id
+  rpc StartPriceFeed(StartPriceFeedRequest) returns (StartPriceFeedResponse);
+  // StopPriceFeed stops the price feed with the given id
+  rpc StopPriceFeed(StopPriceFeedRequest) returns (StopPriceFeedResponse);
+  // UpdatePriceFeed updates the price feed
```
altafan:
```suggestion
  // UpdatePriceFeed allows to change source and/or ticker of the given price feed.
```

api-spec/protobuf/tdex-daemon/v2/feeder.proto
```diff
@@ -0,0 +1,84 @@
+syntax = "proto3";
```
altafan, line 1: NIT: In other protos all comments end with a `.`, we could maintain this convention for the those added

api-spec/protobuf/tdex-daemon/v2/types.proto
```diff
@@ -263,3 +263,16 @@ message CustomPeriod {
...
+  // id is the id of the price feed
+  string id = 1;
+  // market is the market for which the price feed is created
+  tdex.v1.Market market = 2;
+  // source is the name of the price source to use
+  string source = 3;
+  // ticker is the ticker of the asset to use as price source
+  string ticker = 4;
+  // on is the flag to indicate if the price feed is turned on or off
+  bool on = 5;
```
altafan: maybe `started` is more expressive?
```suggestion
  bool  started = 5;
```

internal/core/application/feeder/service.go
```diff
@@ -0,0 +1,176 @@
...
+	ctx context.Context,
+	req ports.AddPriceFeedReq,
+) (string, error) {
+	return s.feederSvc.AddPriceFeed(ctx, req)
+}
+
+func (s *Service) UpdatePriceFeed(
+	ctx context.Context,
+	req ports.UpdatePriceFeedReq,
+) error {
```
altafan, line 124: let's keep things simple for this.. we don't really need an interface as arghere, we can have just:

```suggestion
func (s *Service) UpdatePriceFeed(
	ctx context.Context, id, source, ticker string,
) error {
```

internal/core/application/feeder/service.go
```diff
@@ -0,0 +1,176 @@
...
+	feederSvc   ports.PriceFeeder
+}
+
+func NewService(
+	repoManager ports.RepoManager,
+) Service {
+	return Service{repoManager: repoManager}
+}
+
+func (s *Service) Start(ctx context.Context) error {
```
altafan: we don't have Start/Stop methods in the other app services at the moment.. they start doing their job usually by a `listenToWhatever` function like done [here](https://github.com/tdex-network/tdex-daemon/blob/v1/internal/core/application/wallet/service.go#L43-L44) or by registering handlers for events like [this one](https://github.com/tdex-network/tdex-daemon/blob/v1/internal/core/application/operator/service.go#L62-L63).

Let's keep this approach for now, then we will change all of them in a dedicated PR.
In this case, I don't think we need events so it's enough to add a `listenToPriceFeeds` method or something like that.

Also, you can see here how we need the missing `ListPriceFeeds` API and related ones in other layers (storage) so we can list them all and start only those that are not stopped (instead of all of them) without interacting with any repository.

altafan: The feeder should also be stopped at interface layer as done for [other services](https://github.com/tdex-network/tdex-daemon/blob/v1/internal/interfaces/grpc/service.go#L320-L327). I can only agree with you if you say that this is wrong, but again let's make this kind of change in a dedicated PR to not create any mess.

internal/core/application/feeder/error.go
```diff
@@ -0,0 +1,9 @@
+package feeder
```
altafan: As you can notice, none of the other app services have any hardcoded error and this is actually intended.
My idea was to dedicate a PR (with prior discussion) to have a common place for all the currently repeated internal types, and also constants and errors. Not high priority at this stage tho.
Therefore, I ask you to drop this file and return `fmt.Errorf` directly in the app service methods.

internal/core/application/feeder/service.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+func (s *Service) StopFeed(ctx context.Context, feedID string) error {
+	return s.feederSvc.StopFeed(ctx, feedID)
+}
+
+func (s *Service) AddPriceFeed(
+	ctx context.Context,
+	req ports.AddPriceFeedReq,
+) (string, error) {
```
altafan, line 117: Here we could just represent a price feed to be added as a PriceFeed with empty id and started: false. The goal is to have and use only essential portable types and not a custom ones to fit any specific need.

```suggestion
func (s *Service) AddPriceFeed(
	ctx context.Context, feed ports.PriceFeed,
) (string, error) {
```

altafan: we can have a more "low-level" signature here like:
```suggestion
func (s *Service) AddPriceFeed(
	ctx context.Context, market ports.Market, source, ticker string,
) (string, error) {
```

The goal is to reuse the portable types we already have as much as possible instead of adding custom ones to fit any new need.

internal/core/application/feeder/service.go
altafan: Same here, we can simplify to:
```suggestion
func (s *Service) UpdatePriceFeed(
	ctx context.Context, id, source, ticker string,
) error {
```

internal/core/application/feeder/service.go
```diff
@@ -0,0 +1,176 @@
...
+		return ErrFeedOn
+	}
+
+	return s.feederSvc.UpdatePriceFeed(ctx, req)
+}
+
+func (s *Service) RemovePriceFeed(
+	ctx context.Context,
+	feedID string,
+) error {
```
altafan: In this repo the convention is to write in-line args as long as they are max 80 chars, let's also avoid to repeat words:
```suggestion
func (s *Service) RemovePriceFeed(ctx context.Context, id string) error {
```

internal/core/application/feeder/service.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	return s.feederSvc.RemovePriceFeed(ctx, feedID)
+}
+
+func (s *Service) GetPriceFeed(
+	ctx context.Context,
+	baseAsset string,
+	quoteAsset string,
+) (*ports.PriceFeed, error) {
```
altafan: also here we can write in-line like:
```suggestion
func (s *Service) GetPriceFeed(
	ctx context.Context, baseAsset, quoteAsset string,
) (ports.PriceFeed, error) {
```

internal/core/application/feeder/types.go
```diff
@@ -0,0 +1,13 @@
...
+	baseAsset  string
+	quoteAsset string
+}
+
+func (i marketInfo) GetBaseAsset() string {
+	return i.baseAsset
+}
+func (i marketInfo) GetQuoteAsset() string {
+	return i.quoteAsset
+}
```
altafan: we can simplify this a little bit like:

```suggestion
type marketInfo domain.Market

func (i marketInfo) GetBaseAsset() string {
	return i.BaseAsset
}
func (i marketInfo) GetQuoteAsset() string {
	return i.QuoteAsset
}
```

internal/core/ports/feeder.go
```diff
@@ -0,0 +1,28 @@
...
+import "context"
+
+type PriceFeeder interface {
+	// Start starts the price feeder service which is periodically fetching
+	//and updating market prices from external price provide
+	Start(ctx context.Context, markets []Market) (chan PriceFeedChan, error)
+	// Stop stops the price feeder service
+	Stop(ctx context.Context)
+	// StartFeed starts updating specific price feed for the given feedID
+	StartFeed(ctx context.Context, feedID string) error
```
altafan: I don't remember the choice we made for this project, but in all the codebase there should be only `Id` or `ID`. To shortcut this, let's just use `id` instead of `feedID` in the methods of this interface :)

```suggestion
	// StartFeed starts forwarding price feeds from the given source to the target market.
	StartFeed(ctx context.Context, id string) error
```

internal/core/ports/feeder.go
```diff
@@ -0,0 +1,28 @@
...
+	StopFeed(ctx context.Context, feedID string) error
+	// AddPriceFeed adds a new price feed
+	//it is necessary to start the feed after adding it in order to start updating prices
+	AddPriceFeed(ctx context.Context, req AddPriceFeedReq) (string, error)
+	// UpdatePriceFeed updates an existing price feed
+	UpdatePriceFeed(ctx context.Context, req UpdatePriceFeedReq) error
+	// RemovePriceFeed removes an existing price feed
+	RemovePriceFeed(ctx context.Context, feedID string) error
+	// GetPriceFeed returns the price feed for the given market
+	GetPriceFeed(ctx context.Context, baseAsset string, quoteAsset string) (*PriceFeed, error)
```
altafan: Do we really need this API if we add `ListPriceFeeds()`? If I'm not wrong this is currently used when starting the price feeds. If we still need this we can call it:
```suggestion
	GetPriceFeedForMarket(ctx context.Context, baseAsset string, quoteAsset string) (*PriceFeed, error)
```

internal/core/ports/feeder.go
```diff
@@ -0,0 +1,28 @@
...
+	// GetPriceFeedForFeedID returns the price feed for the given feedID
+	GetPriceFeedForFeedID(ctx context.Context, feedID string) (*PriceFeed, error)
```
altafan: this can become either:
```suggestion
	// GetPriceFeedForID returns the price feed with the given id.
	GetPriceFeedForID(ctx context.Context, id string) (PriceFeed, error)
```

or:

```suggestion
	// GetPriceFeedForFeedID returns the price feed for the given feedID
	GetPriceFeed(ctx context.Context, id string) (PriceFeed, error)
```

depending on the fate of the actual `GetPriceFeed`.

internal/core/ports/types.go
```diff
@@ -252,3 +252,30 @@ type BuildData interface {
...
+	MarketQuoteAsset string
+	Source           string
+	Ticker           string
+}
+
+type UpdatePriceFeedReq struct {
+	ID     string
+	Source string
+	Ticker string
+}
```
altafan: In spike of the previous comments these can be dropped

internal/core/ports/types.go
```diff
@@ -252,3 +252,30 @@ type BuildData interface {
...
+
+type PriceFeed struct {
+	ID               string
+	MarketBaseAsset  string
+	MarketQuoteAsset string
+	Source           string
+	Ticker           string
+	On               bool
+}
```
altafan: As you can notice, there are no structs defined here, only interfaces in order to make them usable from both the app and the interface layers.
We may refactor this, but, as always, for now let's stick with the actual style and eventually refactor in a dedicated PR:
```suggestion
type PriceFeed struct {
	GetId() string
	GetMarket() Market
	GetSource() string
	GetTicker() string
	IsStarted() bool
}
```

internal/infrastructure/price-feeder/repository.go
```diff
@@ -0,0 +1,25 @@
+package pricefeederinfra
+
+import (
+	"context"
+)
+
+type PriceFeedRepository interface {
```
altafan: you're totally right in defining this as repository but I'm afraid it could create confusion with domain repos.

What do you think about calling this?
```suggestion
type PriceFeedStore interface {
```

internal/infrastructure/price-feeder/repository.go
```diff
@@ -0,0 +1,25 @@
...
+	// AddPriceFeed adds a new price feed to the repository.
+	AddPriceFeed(ctx context.Context, priceFeed PriceFeed) error
+	// GetPriceFeed returns the price feed with the given ID.
+	GetPriceFeed(ctx context.Context, id string) (*PriceFeed, error)
+	// GetPriceFeedsByMarket returns all price feed for a given market.
+	GetPriceFeedsByMarket(
```
altafan: ah! you used a naming that i do really like here but you basically reverted in the portable interface by specifying GetPriceFeed and GetPriceFeedForFeedID.

This can cause confusion so let's use this naming also in `ports.PriceFeeder` as long as we still need a way to get price feeds by market.

internal/infrastructure/price-feeder/domain.go
```diff
@@ -0,0 +1,35 @@
+package pricefeederinfra
```
altafan: same for this file, i totally understand the naming but to prevent any kind of confusion, we can call it price_feed.go or feed.go, as you wish.

internal/infrastructure/price-feeder/service.go
```diff
@@ -0,0 +1,265 @@
...
+
+	p.addMarkets([]pricefeeder.Market{
+		{
+			BaseAsset:  priceFeed.Market.BaseAsset,
+			QuoteAsset: priceFeed.Market.QuoteAsset,
+			Ticker:     priceFeed.Market.Ticker,
+		},
+	})
+
+	return p.watchMarkets(p.getMarkets())
```
altafan: I guess this comes from the old implementation where we had a different approach and start all price feeds for all markets defined in a source config file.
Now we are shifting to a totally different one, so here theoretically we should interact with the price source and start receiving feeds from it that will be forwarded to a channel.. on the other side of this channel there will be the app service that will update markets prices.

internal/infrastructure/price-feeder/service.go
```diff
@@ -0,0 +1,265 @@
...
+
+	p.removeMarkets([]pricefeeder.Market{
+		{
+			BaseAsset:  priceFeed.Market.BaseAsset,
+			QuoteAsset: priceFeed.Market.QuoteAsset,
+			Ticker:     priceFeed.Market.Ticker,
+		},
+	})
+
+	return p.watchMarkets(p.getMarkets())
```
altafan: same, not needed.. i do think we don't even need to store markets in-memory at all.

Here, in theory, we should contact the price source and make it stop sending us price feeds for the related market.

sekulicd: Done

internal/infrastructure/price-feeder/service.go
sekulicd: Done

internal/infrastructure/price-feeder/domain.go
sekulicd: Done

internal/infrastructure/price-feeder/repository.go
sekulicd: Done

internal/infrastructure/price-feeder/repository.go
sekulicd: Done

internal/core/ports/types.go
sekulicd: Done

internal/core/ports/types.go
sekulicd: Done

internal/core/ports/feeder.go
sekulicd: Done

internal/core/ports/feeder.go
sekulicd: Done

internal/core/ports/feeder.go
sekulicd: Done

internal/core/application/feeder/types.go
sekulicd: Done

internal/core/application/feeder/service.go
sekulicd: Done

internal/core/application/feeder/service.go
sekulicd: Done

internal/core/application/feeder/service.go
sekulicd: Done

internal/core/application/feeder/service.go
sekulicd: Done

internal/core/application/feeder/error.go
sekulicd: Done

internal/core/application/feeder/service.go
sekulicd: Done

api-spec/protobuf/tdex-daemon/v2/types.proto
sekulicd: Done

api-spec/protobuf/tdex-daemon/v2/feeder.proto
sekulicd: Done

api-spec/protobuf/tdex-daemon/v2/feeder.proto
sekulicd: Done

api-spec/protobuf/tdex-daemon/v2/feeder.proto
sekulicd: Done

cmd/tdex/feeder.go
```diff
@@ -0,0 +1,305 @@
...
+	defer cleanup()
+
+	baseAsset := ctx.String("base_asset")
+	quoteAsset := ctx.String("quote_asset")
+	source := ctx.String("source")
+	ticker := ctx.String("ticker")
+
+	if baseAsset == "" || quoteAsset == "" || source == "" || ticker == "" {
+		return cli.Exit("base_asset, quote_asset, source and ticker are required", 1)
+	}
```
altafan: not needed, we can set `Required: true` for the relative flags like:
```
                        &cli.StringFlag{
				Name:  "base_asset",
				Usage: "base asset of the market for which the price feed is created",
                                Required: true,
			},
```

cmd/tdex/feeder.go
```diff
@@ -0,0 +1,305 @@
...
+
+	if _, err := client.AddPriceFeed(ctx.Context, &daemonv2.AddPriceFeedRequest{
+		Market: &tdexv2.Market{
+			BaseAsset:  baseAsset,
+			QuoteAsset: quoteAsset,
+		},
+		Source: source,
+		Ticker: ticker,
+	}); err != nil {
+		return cli.Exit(err, 1)
```
altafan: never used `cli.Exit(err, 1)` in other commands, let's keep returning an error as always (please make this change wherever required):
```suggestion
		return err
```

cmd/tdex/feeder.go
altafan: Please make the change wherever required

cmd/tdex/feeder.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+
+	fmt.Println("Price feed added")
```
altafan, line 150: Messages must not be capitalized, please make this change wherever required.

altafan: Let's print the price feed id here, similar to what we do when we [add a weboohk](https://github.com/tdex-network/tdex-daemon/blob/v1/cmd/tdex/addwebhook.go#L59-L60)

Also, in general, messages must not be capitalized as per golang convention, please make this change wherever required
```suggestion
        fmt.Println() 
	fmt.Println("price feed id:",  reply.GetId())
```

cmd/tdex/feeder.go
```diff
@@ -0,0 +1,305 @@
...
+		return err
+	}
+	defer cleanup()
+
+	id := ctx.String("id")
+	source := ctx.String("source")
+	ticker := ctx.String("ticker")
+	if id == "" || source == "" || ticker == "" {
+		return cli.Exit("id, source and ticker are required", 1)
+	}
```
altafan: in this specific command only `id` must be required, `source` and `ticker` are optional instead, since the operator might want to change only the ticker because he used the wrong one for example.

Also, it should be responsibility of the daemon to return an error in case the operator passes empty source and ticker in the message request. At least one of the two must not be empty. This check should be done at infra level I think.

cmd/tdex/feeder.go
```diff
@@ -0,0 +1,305 @@
...
+}
+
+func listPriceFeedsAction(ctx *cli.Context) error {
+	client, cleanup, err := getFeederClient(ctx)
+	if err != nil {
+		return err
+	}
+	defer cleanup()
+
+	reply, err := client.ListPriceFeeds(ctx.Context, &daemonv2.ListPriceFeedsRequest{})
```
altafan:
```suggestion
	reply, err := client.ListPriceFeeds(
	   ctx.Context, &daemonv2.ListPriceFeedsRequest{},
	)
```

cmd/tdex/feeder.go
```diff
@@ -0,0 +1,305 @@
...
+
+func listSourcesAction(ctx *cli.Context) error {
+	client, cleanup, err := getFeederClient(ctx)
+	if err != nil {
+		return err
+	}
+	defer cleanup()
+
+	reply, err := client.ListSupportedPriceSources(ctx.Context,
+		&daemonv2.ListSupportedPriceSourcesRequest{})
```
altafan, line 281:
```suggestion
	reply, err := client.ListSupportedPriceSources(
	        ctx.Context, &daemonv2.ListSupportedPriceSourcesRequest{},
	)
```

cmd/tdex/feeder.go
```diff
@@ -0,0 +1,305 @@
...
+		},
+	}
+	updatePriceFeed = &cli.Command{
+		Name:   "update",
+		Usage:  "update a price feed source and ticker",
+		Action: updatePriceFeedAction,
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:  "id",
+				Usage: "id of the price feed to start",
```
altafan:
```suggestion
				Usage: "id of the price feed to update",
```

cmd/tdex/feeder.go
```diff
@@ -0,0 +1,305 @@
...
+		},
+	}
+	removePriceFeed = &cli.Command{
+		Name:   "remove",
+		Usage:  "remove a price feed",
+		Action: removePriceFeedAction,
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:  "id",
+				Usage: "id of the price feed to start",
```
altafan:
```suggestion
				Usage: "id of the price feed to remove",
```

cmd/tdex/feeder.go
```diff
@@ -0,0 +1,305 @@
...
+		Action: addPriceFeedAction,
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:  "base_asset",
+				Usage: "base asset of the market for which the price feed is created",
+			},
+			&cli.StringFlag{
+				Name:  "quote_asset",
+				Usage: "quote asset of the market for which the price feed is created",
+			},
```
altafan: In other commands, we source the market from the CLI state instead of making use of flags to define market's base/quote asset, like done for example for the [`marketfragmenter split`](https://github.com/tdex-network/tdex-daemon/blob/v1/cmd/tdex/marketfragmenter.go#L143-L146) one.

Let's stick with this approach for the moment, therefore we can drop base/quote asset flags from command.

Also, just to make you aware of this, we are planning to refer to a market with label or univoque id instead of its asset pair. The ticker is yet to be opened though.

api-spec/protobuf/tdex-daemon/v2/feeder.proto
```diff
@@ -0,0 +1,93 @@
...
+message ListSupportedPriceSourcesRequest {}
+
+message ListSupportedPriceSourcesResponse {
+  // sources is the list of supported price sources.
+  repeated string sources = 1;
+}
+
+message GetPriceFeedRequest {
+  // market is the market for which the price feeds are listed.
+  tdex.v2.Market market = 1;
```
altafan: We refer to a price feed with `id` in all other RPCs, don't see why we should do something different here:
```suggestion
  string id = 1;
```

cmd/tdex/feeder.go
```diff
@@ -0,0 +1,305 @@
...
+		},
```
altafan: In spike of the prev comments, these flags can be dropped in favour of a `--id` flag:
```suggestion
			&cli.StringFlag{
				Name:  "id",
				Usage: "id of the price feed to get info about",
				Required: true,
			},
```

cmd/tdex/feeder.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+	stopPriceFeed = &cli.Command{
+		Name:   "stop",
+		Usage:  "stops price feed",
+		Action: stopPriceFeedAction,
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:  "id",
+				Usage: "id of the price feed to start",
```
altafan:
```suggestion
				Usage: "id of the price feed to stop",
```

internal/core/application/feeder.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"context"
+
+	"github.com/tdex-network/tdex-daemon/internal/core/application/feeder"
+
+	"github.com/tdex-network/tdex-daemon/internal/core/ports"
+)
+
+type FeederService interface {
+	Stop(ctx context.Context)
```
altafan, line 12: What's this used for?

internal/core/application/feeder/service.go
```diff
@@ -0,0 +1,175 @@
...
+	if err := svc.startPriceFeed(feederSvc); err != nil {
+		return Service{}, err
+	}
+
+	return svc, nil
+}
+
+func (s *Service) Stop(ctx context.Context) {
+	s.feederSvc.Stop(ctx)
+}
```
altafan: This shouldn't be needed actually since the same service is already stopped at interface layer. Can we remove this method?

internal/core/application/feeder/service.go
```diff
@@ -0,0 +1,175 @@
...
+
+func (s *Service) StopFeed(ctx context.Context, feedID string) error {
+	return s.feederSvc.StopFeed(ctx, feedID)
+}
+
+func (s *Service) AddPriceFeed(
+	ctx context.Context, market ports.Market, source, ticker string,
+) (string, error) {
+	return s.feederSvc.AddPriceFeed(ctx, market, source, ticker)
+}
```
altafan: let's check if the market really exists:
```suggestion
) (string, error) {
        if _, err := s.repoManager.MarketRepository().GetMarketByAssets(
		ctx, market.GetBaseAsset(), market.GetQuoteAsset(),
	); err != nil {
	        return "", err
	}
	return s.feederSvc.AddPriceFeed(ctx, market, source, ticker)
}
```

internal/core/application/feeder/service.go
```diff
@@ -0,0 +1,175 @@
...
+func (s *Service) UpdatePriceFeed(
+	ctx context.Context, id, source, ticker string,
+) error {
+	priceFeed, err := s.feederSvc.GetPriceFeed(ctx, id)
+	if err != nil {
+		return err
+	}
+
+	if priceFeed.IsStarted() {
+		return fmt.Errorf("feed needs to be stopped before updating it")
```
altafan:
```suggestion
		return fmt.Errorf("price feed must be stopped to be updated")
```

internal/core/application/feeder/service.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+func (s *Service) RemovePriceFeed(ctx context.Context, id string) error {
+	priceFeed, err := s.feederSvc.GetPriceFeed(ctx, id)
+	if err != nil {
+		return err
+	}
+
+	if priceFeed.IsStarted() {
+		return fmt.Errorf("feed needs to be stopped before updating it")
```
altafan:
```suggestion
		return fmt.Errorf("price feed must be stopped to be removed")
```

internal/core/application/feeder/service.go
```diff
@@ -0,0 +1,175 @@
...
+		ctx,
+		priceFeed.GetMarket().GetBaseAsset(),
+		priceFeed.GetMarket().GetQuoteAsset(),
+	)
+	if err != nil {
+		return err
+	}
+
+	if market.IsStrategyPluggable() {
+		return fmt.Errorf("cant remove pluggable market")
```
altafan:
```suggestion
		return fmt.Errorf("market strategy must not be pluggable to remove price feed")
```

internal/core/application/feeder/service.go
```diff
@@ -0,0 +1,175 @@
...
+
+func (s *Service) ListSources(ctx context.Context) []string {
+	return s.feederSvc.ListSources(ctx)
+}
+
+func (s *Service) ListPriceFeeds(ctx context.Context) ([]ports.PriceFeed, error) {
+	return s.feederSvc.ListPriceFeeds(ctx)
+}
+
+func (s *Service) listenPriceFeedChan(priceFeedChan chan ports.PriceFeedChan) {
```
altafan: Not sure having a single channel for all price feeds is the best design for this feature mostly because the daemon this way starts a go routine to listen to a channel that potentially can never be used by the operator, if, for example, he never sets up a price feed for any market. Sounds a bit like wasting resources somehow.

I do think it would be better having a dedicated channel for every price feed, this way we could:
- start a go routine to listen to a channel only once the price feed is started
- stop a go routine that listens to the channel right after the price feed is stopped

With this design, at app level we end up having a go routine for each active price feed. If there are no price feeds, the daemon doesn't perform any background job at all.

internal/infrastructure/price-feeder/feed.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func (p PriceFeed) IsStarted() bool {
+	return p.Started
+}
+
+type Market struct {
+	BaseAsset  string
+	QuoteAsset string
+	Ticker     string
```
altafan: Let's move ticker to `PriceFeed` like the proto.

internal/infrastructure/price-feeder/feed.go
```diff
@@ -0,0 +1,56 @@
+package pricefeederinfra
```
altafan, line 1: Are we sure we need to export the types declared here? If not needed, we can move to `priceFeed` and `market` and make all their fields not exported too.

Also, if the factory func is not used outside of this folder, we should make it internal: `newPriceFeed`

internal/core/application/feeder/service.go
sekulicd: If concern are resources in case of un-used chan i dont see this as issue, cause this is insignificant impact, initial go routine alloc is ~2KB.
If the concern is processing speed(and potential wait on chan "unlock") of price feed when received on chan we can put db update in go routine(talking about code in [here](https://github.com/sekulicd/tdex-daemon/blob/cfa1ec5dbaf46930f2f6df7a4f39c99bd55ca795/internal/core/application/feeder/service.go#L130) and maybe update to buffered chan.
Adding multiple chans would add more complexity to already bit complex design and i think redesign of both pkg's and infra feeder would be needed.
My suggestion, is to continue with current approach, i tested it pretty solid and seems like working fine, if we want to optimise for performance or reason about new design maybe its better to move to another issue as this is already big PR.

cmd/tdex/feeder.go
sekulicd: Done

cmd/tdex/feeder.go
sekulicd: Done

cmd/tdex/feeder.go
sekulicd: Done

cmd/tdex/feeder.go
sekulicd: Done

cmd/tdex/feeder.go
sekulicd: Done

cmd/tdex/feeder.go
sekulicd: Done

cmd/tdex/feeder.go
sekulicd: Done

cmd/tdex/feeder.go
sekulicd: Done

cmd/tdex/feeder.go
sekulicd: Done

api-spec/protobuf/tdex-daemon/v2/feeder.proto
sekulicd: Done

cmd/tdex/feeder.go
sekulicd: Done

cmd/tdex/feeder.go
sekulicd: Done

internal/core/application/feeder/service.go
sekulicd: I dont get u, can u elaborate on this? @altafan

internal/core/application/feeder/service.go
sekulicd: Done

internal/core/application/feeder/service.go
sekulicd: Done

internal/core/application/feeder/service.go
sekulicd: Done

internal/core/application/feeder/service.go
sekulicd: Done

internal/infrastructure/price-feeder/feed.go
sekulicd: Done

altafan (approved):

altafan:
Closes #657.

PR #701, opened 2023-04-19
Update ocean protos

altafan:
This updates ocean's protos to their latest version, where accounts now have a namespace and label. From operator's POV this doesn't mean that much in the sense that there's no change at interface level beside changes to WalletInfo and AccountInfo types according to ocean (the master blinding key has been moved from the former to the latter).

Internally, instead, it's been necessary adding a way to reference an account label and namespace to correctly identify accounts since ocean refers to them always by namespace.

NOTE: The way this is designed implies that every wallet account must have a name (label) that cannot be changed later even if this is allowed by ocean. This is because tdex-daemon uses account labels to identify them but also as reference for other domains and changing a label would require making the proper change in every domain, which is not sustainable.

Please @sekulicd review.


sekulicd (approved):
LGTM

PR #708, opened 2023-05-02
Add strategy to NewMarket & Fix typo

### LINKED ISSUE #706: [v1] Add MarketStrategy to NewMarket request
NewMarket should take an additional MarketStrategy parameter so that client can set it in market creation flow.
If Pluggable is chosen then price should be set to zero.

altafan:
This allows the operator to define the market strategy at creation time.

Closes #706.

Please @sekulicd review this.


internal/interfaces/grpc/handler/utils.go
```diff
@@ -57,14 +57,25 @@ func parseMarketFee(fee *tdexv2.MarketFee) (int64, int64, error) {
 	return baseFee, quoteFee, nil
 }
 
+func parseStrategySafe(strategyType daemonv2.StrategyType) (int, error) {
```
sekulicd, line 60: can we have one method that is parsing StrategyType?

internal/interfaces/grpc/handler/utils.go
```diff
@@ -57,14 +57,25 @@ func parseMarketFee(fee *tdexv2.MarketFee) (int64, int64, error) {
...
+	strategy, err := parseStrategy(strategyType)
+	if err != nil {
+		return -1, err
+	}
+	if strategy == domain.StrategyTypeUndefined {
```
sekulicd, line 65: case domain. StrategyTypeUndefined is never returned by the parseStrategy function?

internal/core/domain/market.go
```diff
@@ -104,6 +104,12 @@ func NewMarket(
 	if !isValidPrecision(quoteAssetPrecision) {
 		return nil, ErrMarketInvalidQuoteAssetPrecision
 	}
+	if !isValidStrategy(strategyType) {
+		return nil, ErrMarketUnknownStrategy
+	}
+	if strategyType == StrategyTypeUndefined {
+		strategyType = StrategyTypeBalanced
```
sekulicd, line 111: here u set type to StrategyTypeBalanced if "undefined/unspecified" and in handler/utils.go if it is StrategyType_STRATEGY_TYPE_UNSPECIFIED u set type to StrategyTypePluggable?

internal/interfaces/grpc/handler/utils.go
altafan: there was already an exisitng `parseStrategy` function taking care of converting a StrategyType used by `UpdateMarketStrategy` RPC. In this RPC `StrategyTypeUndefined` is not a valid strategy, while it is in `NewMarket` (we default to balanced strategy if it is not specified by the user)

internal/interfaces/grpc/handler/utils.go
altafan: https://github.com/tdex-network/tdex-daemon/pull/708#discussion_r1183576891

internal/core/domain/market.go
altafan: https://github.com/tdex-network/tdex-daemon/pull/708#discussion_r1183576891

sekulicd (approved):

PR #709, opened 2023-05-03
Health check

### LINKED ISSUE #707: [v1] HealthCheck endpoint

sekulicd:
This adds health check endpoint.

This closes #707 

@altafan please review.


altafan:
@sekulicd let's add the healthcheck handler to the operator service too.

Also, we should update permissions_test.go to make it import health proto and test that all its methods are in the whitelist as done for other protos.

sekulicd:
> @sekulicd let's add the healthcheck handler to the operator service too.
> 
> Also, we should update permissions_test.go to make it import health proto and test that all its methods are in the whitelist as done for other protos.

Done

internal/interfaces/grpc/permissions/permissions_test.go
```diff
@@ -34,6 +36,10 @@ func TestWhitelistedMethods(t *testing.T) {
 		allMethods = append(allMethods, fmt.Sprintf("/%s/%s", tdexv2.TradeService_ServiceDesc.ServiceName, v.MethodName))
 	}
 
+	for range grpchealth.Health_ServiceDesc.ServiceName {
+		allMethods = append(allMethods, fmt.Sprintf("/%s/%s", grpchealth.Health_ServiceDesc.ServiceName, "Check"))
+	}
```
altafan:
```suggestion
	for _, v := range grpchealth.Health_ServiceDesc.ServiceName {
		allMethods = append(allMethods, fmt.Sprintf("/%s/%s", grpchealth.Health_ServiceDesc.ServiceName, v.MethodName))
	}
```

This implies adding Watch to the whitelist even if it's not implemented basically

sekulicd: Is it better if we dont add it to whitelist cause it is not implemented?

altafan: alright, so this should change like: 
```suggestion
	allMethods = append(allMethods, fmt.Sprintf("/%s/%s", grpchealth.Health_ServiceDesc.ServiceName, "Check"))
```

altafan (approved):

PR #710, opened 2023-05-04
Postgres database

sekulicd:
This adds pg db implementation.

@altafan please review.


sekulicd:
@altafan please review proposed pg db schema.

internal/infrastructure/storage/db/pg/migration/20230504102232_init.up.sql
```diff
@@ -0,0 +1,63 @@
...
+);
+
+CREATE TABLE fee (
+    id SERIAL PRIMARY KEY,
+    base_asset_fee BIGINT NOT NULL,
+    quote_asset_fee BIGINT NOT NULL,
+    type VARCHAR(10) NOT NULL CHECK (type IN ('percentage', 'fixed')),
+    fk_market_name VARCHAR(50) NOT NULL,
+    FOREIGN KEY (fk_market_name) REFERENCES market (name)
+);
```
altafan: Maybe better to store base/quote asset fee directly in the market table as fields `percentage_base_fee | percentage_quote_fee | fixed_base_fee | fixed_quote_fee`?

internal/infrastructure/storage/db/pg/migration/20230504102232_init.up.sql
```diff
@@ -0,0 +1,63 @@
+CREATE TABLE market (
+    name VARCHAR(50) NOT NULL PRIMARY KEY,
```
altafan, line 2:
```suggestion
    name VARCHAR(255) NOT NULL PRIMARY KEY,
```

internal/infrastructure/storage/db/pg/migration/20230504102232_init.up.sql
```diff
@@ -0,0 +1,63 @@
...
+    fk_market_name VARCHAR(50) NOT NULL,
+    FOREIGN KEY (fk_market_name) REFERENCES market (name)
+);
+
+CREATE TABLE trade (
+    id VARCHAR(50) PRIMARY KEY,
+    type INTEGER NOT NULL,
+    fee_asset VARCHAR(255) NOT NULL,
+    fee_amount BIGINT NOT NULL,
+    trader_pubkey BYTEA NOT NULL,
```
altafan: This is actualy not used at app level, so it should be nullable:

```suggestion
    trader_pubkey BYTEA,
```

internal/infrastructure/storage/db/pg/migration/20230504102232_init.up.sql
altafan, line 2: This implies that in other layers (domain, interface) we MUST validate the name specified by the user and check that length is <= 50.

internal/infrastructure/storage/db/pg/migration/20230504102232_init.up.sql
```diff
@@ -0,0 +1,63 @@
+CREATE TABLE market (
+    name VARCHAR(50) NOT NULL PRIMARY KEY,
+    base_asset VARCHAR(255) NOT NULL,
+    quote_asset VARCHAR(255) NOT NULL,
```
altafan:
```suggestion
    base_asset VARCHAR(64) NOT NULL,
    quote_asset VARCHAR(64) NOT NULL,
```

internal/infrastructure/storage/db/pg/migration/20230504102232_init.up.sql
```diff
@@ -0,0 +1,63 @@
...
+    base_asset_precision INTEGER NOT NULL DEFAULT 8,
+    quote_asset_precision INTEGER NOT NULL DEFAULT 8,
```
altafan: Remove defaults here, they are handled at domain level - actually, the default value used at domain level is 0.
```suggestion
    base_asset_precision INTEGER NOT NULL,
    quote_asset_precision INTEGER NOT NULL,
```

internal/infrastructure/storage/db/pg/migration/20230504102232_init.up.sql
```diff
@@ -0,0 +1,63 @@
...
+    tradeable BOOLEAN NOT NULL DEFAULT FALSE,
```
altafan: This is not really a typo, but the field at domain level is called Tradable, so let's stick with it. Also, we can remove defaults.

```suggestion
    tradable BOOLEAN NOT NULL,
```

internal/infrastructure/storage/db/pg/migration/20230504102232_init.up.sql
```diff
@@ -0,0 +1,63 @@
...
+    strategy_type INTEGER NOT NULL DEFAULT 0,
+    base_price INTEGER NOT NULL DEFAULT 0,
+    quote_price INTEGER NOT NULL DEFAULT 0,
```
altafan: Defaults are not needed as the domain doesn't expect these fields to be nil:

```suggestion
    strategy_type INTEGER NOT,
    base_price INTEGER NOT NULL,
    quote_price INTEGER NOT NULL,
```

altafan: note that in badger impl we decided to split market and price files as prices might be constantly updated by a feeder and the main reason was preventing conflicts in case the operator and the feeder updated the same market at nearly the same time.

Shall we have a dedicated table for prices in this pg impl? I'm not really concerned about conflicts here, but the fact that prices can be subject to multiple changes make me thinking that would be better to isolate them.

internal/infrastructure/storage/db/pg/migration/20230504102232_init.up.sql
```diff
@@ -0,0 +1,63 @@
...
+);
+
+CREATE TABLE trade (
+    id VARCHAR(50) PRIMARY KEY,
+    type INTEGER NOT NULL,
+    fee_asset VARCHAR(255) NOT NULL,
+    fee_amount BIGINT NOT NULL,
+    trader_pubkey BYTEA NOT NULL,
+    status_code INTEGER NOT NULL DEFAULT 0,
+    status_failed BOOLEAN NOT NULL DEFAULT FALSE,
```
altafan: remove defaults:
```suggestion
    status_code INTEGER NOT NULL,
    status_failed BOOLEAN NOT NULL,
```

internal/infrastructure/storage/db/pg/migration/20230504102232_init.up.sql
```diff
@@ -0,0 +1,63 @@
...
+    pset_base64 TEXT NOT NULL,
+    tx_id VARCHAR(255) NOT NULL,
+    tx_hex TEXT NOT NULL,
+    expiry_time BIGINT NOT NULL,
+    settlement_time BIGINT NOT NULL,
```
altafan: All these fields can be empty actaully.. shall we declare them nullable?

```suggestion
    tx_id VARCHAR(255),
    tx_hex TEXT,
    expiry_time BIGINT,
    settlement_time BIGINT,
```

internal/infrastructure/storage/db/pg/migration/20230504102232_init.up.sql
sekulicd: since we basically have same fields that differs in type(fixed, percentage) i thought it is nicer to have it in separate table which is in 1:M relation to the market table.

internal/infrastructure/storage/db/pg/migration/20230504102232_init.up.sql
```diff
@@ -0,0 +1,63 @@
...
+    timestamp BIGINT NOT NULL,
+    type VARCHAR(10) NOT NULL CHECK (type IN ('request', 'accept', 'complete', 'fail')),
+    fk_trade_id VARCHAR(50) NOT NULL,
+    FOREIGN KEY (fk_trade_id) REFERENCES trade (id)
+);
+
+CREATE TABLE transaction (
+    id SERIAL PRIMARY KEY,
+    type VARCHAR(10) NOT NULL CHECK (type IN ('withdrawal', 'deposit')),
+    account_name TEXT NOT NULL,
```
altafan:
```suggestion
    account_name VARCHAR(50) NOT NULL,
```

internal/infrastructure/storage/db/pg/migration/20230504102232_init.up.sql
```diff
@@ -0,0 +1,63 @@
...
+    tx_id TEXT NOT NULL,
```
altafan:
```suggestion
    txid VARCHAR(64) NOT NULL,
```

internal/infrastructure/storage/db/pg/migration/20230504102232_init.up.sql
```diff
@@ -0,0 +1,63 @@
...
+    timestamp BIGINT NOT NULL
+);
+
+CREATE TABLE transaction_asset_amount (
+    id SERIAL PRIMARY KEY,
+    transaction_id INTEGER NOT NULL,
+    asset TEXT NOT NULL,
```
altafan:
```suggestion
    asset VARCHAR(64) NOT NULL,
```

internal/infrastructure/storage/db/pg/migration/20230504102232_init.up.sql
```diff
@@ -0,0 +1,63 @@
...
+    id SERIAL PRIMARY KEY,
+    type VARCHAR(10) NOT NULL CHECK (type IN ('withdrawal', 'deposit')),
+    account_name TEXT NOT NULL,
+    tx_id TEXT NOT NULL,
+    timestamp BIGINT NOT NULL
+);
+
+CREATE TABLE transaction_asset_amount (
+    id SERIAL PRIMARY KEY,
+    transaction_id INTEGER NOT NULL,
```
altafan: being this a foreign key, should be called:
```suggestion
    fk_transaction_id INTEGER NOT NULL,
```

internal/infrastructure/storage/db/pg/migration/20230504102232_init.up.sql
```diff
@@ -0,0 +1,63 @@
...
+    tx_id VARCHAR(255) NOT NULL,
+    tx_hex TEXT NOT NULL,
+    expiry_time BIGINT NOT NULL,
+    settlement_time BIGINT NOT NULL,
+    fk_market_name VARCHAR(50) NOT NULL,
+    FOREIGN KEY (fk_market_name) REFERENCES market (name)
+);
+
+CREATE TABLE swap (
+    id TEXT PRIMARY KEY,
```
altafan: maybe better? 
```suggestion
    id VARCHAR(255) PRIMARY KEY,
```

internal/infrastructure/storage/db/pg/migration/20230504102232_init.up.sql
altafan: is there a way to enforce that every record in market table must have fixed and percentage fee in this fee table?

altafan: during the review, I saw the same thing done for swaps, no need to change this.

internal/infrastructure/storage/db/pg/migration/20230504102232_init.up.sql
sekulicd: no need to have separate table as it would be 1:1 relation, as for conflicts pg has "row lock" mechanism for updates.

internal/infrastructure/storage/db/pg/sqlc/query.sql
```diff
@@ -0,0 +1,137 @@
...
+-- name: GetAllMarkets :many
+SELECT * FROM market m inner join fee f on m.name = f.fk_market_name;
+
+-- name: UpdateMarket :one
+UPDATE market SET base_asset_precision = $1, quote_asset_precision = $2,
+tradable = $3, strategy_type = $4, base_price = $5, quote_price = $6 WHERE
+name = $7 RETURNING *;
+
+-- name: DeleteMarket :exec
+DELETE FROM market WHERE name = $1;
```
altafan: Are the related fees deleted as well this way?

sekulicd: not with that only, but yes in repo method, check in [here](https://github.com/sekulicd/tdex-daemon/blob/bbdc7ccfe53efd7c998d54a24bd8db2f3ac5da26/internal/infrastructure/storage/db/pg/market_memory_repository_impl.go#L289)

altafan (changes requested):
@sekulicd i noticed that the trade table doesn't contain any market-related field, but rather there's a relation between market and trade tables to retrieve them when querying for trades.

There are a bunch of problems with this approach:
* if a market is deleted, the trades made against it will forever miss market-related info. This can be fixed by deleting markets logically instead of physically by adding an `active` bool field that defaults to true. 
* market fees and price must be stored in the trade table because they represent fees and prices at the time the trade took place. Instead, by keeping a relation between the 2 tables, we might return incorrect values, since price and fees can be subject to changes (in particular the price!) during the life of a market.

internal/infrastructure/storage/db/pg/migration/20230504102232_init.up.sql
```diff
@@ -0,0 +1,64 @@
+CREATE TABLE market (
+    name VARCHAR(50) NOT NULL PRIMARY KEY,
+    base_asset VARCHAR(64) NOT NULL,
+    quote_asset VARCHAR(64) NOT NULL,
+    base_asset_precision INTEGER,
+    quote_asset_precision INTEGER,
```
altafan, line 6: These fields don't need a default but must not be null.

```suggestion
    base_asset_precision INTEGER NOT NULL,
    quote_asset_precision INTEGER NOT NULL,
```

internal/infrastructure/storage/db/pg/migration/20230504102232_init.up.sql
```diff
@@ -0,0 +1,64 @@
...
+    tradable BOOLEAN,
+    strategy_type INTEGER,
+    base_price DOUBLE PRECISION,
+    quote_price DOUBLE PRECISION,
```
altafan: These fields don't need a default but must not be null.
```suggestion
    base_asset_precision INTEGER NOT NULL,
    quote_asset_precision INTEGER NOT NULL,
    tradable BOOLEAN NOT NULL,
    strategy_type INTEGER NOT NULL,
    base_price DOUBLE PRECISION NOT NULL,
    quote_price DOUBLE PRECISION NOT NULL,
```

internal/infrastructure/storage/db/pg/migration/20230504102232_init.up.sql
```diff
@@ -0,0 +1,64 @@
...
+);
+
+CREATE TABLE trade (
+    id VARCHAR(50) PRIMARY KEY,
+    type INTEGER NOT NULL,
+    fee_asset VARCHAR(255) NOT NULL,
+    fee_amount BIGINT NOT NULL,
+    trader_pubkey BYTEA,
+    status_code INTEGER,
+    status_failed BOOLEAN,
```
altafan: these fields don't need defaults but they must be not null.
```suggestion
    status_code INTEGER NOT NULL,
    status_failed BOOLEAN NOT NULL,
```

internal/infrastructure/storage/db/pg/sqlc/query.sql
altafan: since we opted for locigal deletion of markets, this is the best fit for out needs.

A suggestion for the future for the physical deletion scenario, better to let pg handle this by adding a `DELETE ON CASCADE` to the primary key of the table used as fk in others.

internal/infrastructure/storage/db/pg/db_manager.go
```diff
@@ -0,0 +1,174 @@
...
+	pgxPool *pgxpool.Pool
+	querier *queries.Queries
+
+	marketRepository     domain.MarketRepository
+	tradeRepository      domain.TradeRepository
+	depositRepository    domain.DepositRepository
+	withdrawalRepository domain.WithdrawalRepository
+}
+
+func NewService(dbConfig DbConfig) (ports.RepoManager, error) {
```
altafan, line 38: to make things much simpler, we can expect here 2 strings, one being the pg connect URL in the formal `postgresql://user:password@host:port/name` and the other being the migration source path.

This way we can expose just 2 env vars at config level `TDEX_PG_CONNECT_ADDR` and `TDEX_PG_MIGRATION_SOURCE`. This of course implies validating the pg connect addr, which I think it's fine to be done here in this method.

sekulicd: @altafan are u sure about this? in all project we used env var for each db info, changing this also changes out deployment habits a bit

altafan: Yes, definitely better to use 1 env var rather than 4. In the end, the SDK uses the string URL too instead of 4 distinct params. I'd like to make the same changes in other projects too to simplify the configuration.

PR #718, opened 2023-06-08
V.0.9.1 to V1 migration poc

sekulicd:
This adds POC for migrating from v0.9.1 to v1.
Beside migration tls, macaroons, stats, webhooks which should be straightforward, migrating db is bit more complicated.
This is POC that shows how i think it should be done.
Steps:
1. Start v.0.9.1 daemon, create data (wallet, markets, transactions etc)
2. Take datadir as a source for migration CMD
3. Take domain code from v1
4. Translate v.0.9.1 domain to v1 (in this POC Vault to Wallet)

Basically all domains differs from v.0.9.1 to v1 and we need to see how we should map specific fields(we need to make transition map for each domain), i didnt know how to map all of them even for Vault -> Wallet transition, please check in [here](https://github.com/sekulicd/tdex-daemon/blob/1210bb51a18fa3af9cec2095398fd6fe46d70bcb/cmd/migration/v0.9.1-v1/v1-domain/wallet_domain.go#L123).

@tiero @altafan please review and let me know if I should continue in this direction or u have other ideas in mind.


tiero:
Would migration Vault (tdex) to Wallet (ocean) allow the operator to not enter the mnemonic again? If not, let's only migrate tdex v1 domains and teach operator to restore ocean wallet and simply delete/hide current Vault domain.

sekulicd:
> Would migration Vault (tdex) to Wallet (ocean) allow the operator to not enter the mnemonic again? If not, let's only migrate tdex v1 domains and teach operator to restore ocean wallet and simply delete/hide current Vault domain.

Isnt idea to migrate all? Why operator needs to enter mnemonic again?

altafan:
> Would migration Vault (tdex) to Wallet (ocean) allow the operator to not enter the mnemonic again? If not, let's only migrate tdex v1 domains and teach operator to restore ocean wallet and simply delete/hide current Vault domain.

The purpose is to prevent the operator to restore the wallet indeed! at this stage i0m not sure if we can just migrate the db by keeping the wallet locked, otherwise we might need the operator to provide us the password, but we shouldn't need nothing more than that eventually

tiero:
> need the operator to provide us the password

This makes sense, after an updated you used to unlock with password usually, so start migration after that steps seems ok

altafan:
@sekulicd this feature will be merged on `v1` branch, so in cmd we should "swap" the situation, ie. having a folder with all v0.9.1 code to read the old datadir, while importing directly v1 stuff for the translation.

altafan:
Also, to prevent introducing errors, I would expect this folder to contain essentially the "old" domains and badger repo manager to read the data from datadir. We might need to isolate this folder in a dedicated submodule to prevent conflicts between "old" deps and "new" ones (we use badger v2 in v0.x, while we moved to v3 in daemon v1, just to give an example).

Let's go step by step and focus on migrating the main db, we'll take care of other stuff like the webhooks once this main task is done.

PR #720, opened 2023-06-15
Support for version migration

### LINKED ISSUE #715: Migration from v0.9.1 to v1
We currently miss a way to migrate an existing provider `v0.9.1` to the new `v1`.

At the moment this can be partially achieved by starting up the required services and interacting with them to restore the wallet and markets. Trade infos are not restored in this process.

But this is not really a migration from one version to another, because it is done at runtime, so it's more a "start clean and restore what you can".

The purpose here, instead, is to "translate" all the data stored in the "old" format to the new one and basically prepare the datadirs of the services to be started in v1 (ocean, tdexd).

My proposal is to implement a standalone binary like `cmd/migrations` that reads a `v0.9.1`  datadir, translates everything in the new format, cleans the datadir and adds 2 new folders which are the ocean's datadir - with all wallet data - and tdex-daemon's one - with all metadata like markets, fees, prices, trades, etc.

Ideally, once this service finishes its job, it should be enough for the operator to start `oceand` and `tdexd` and use the datadirs created by the migration binary.

From the operator's POV, this operation should look very simple while under the hood it makes quite complex stuff actually.
I think this could be the right approach but I want your feedback guys @sekulicd @tiero.

## Considerations

- **Ocean (but soon also tdexd) can be configured with many db types (pg, badger), should migration support all of them?** We can, like we could allow the operator configuring this binary to write the ocean's data within a datadir or a running pg instance, but to keep things easy at the beginning, and considering that the "old" tdexd uses badger, we could support only migration from "old" to "new" badger and force the operator to use it as ocean's db type.
- **Why a standalone binary? Why not doing this automatically when starting up the v1 daemon the very first time?** Because tdexd requires a running ocean wallet when is started, therefore the migration must be done before starting the services. NOTE: a downside of the standalone binary approach is that in the dockerized solution, one has to override the entrypoint of the daemon container and make it run the migration script instead of the typical one that starts the service, I don't see other option at the moment though.
- **How do we handle conflicts between dependency versions (e.g. v0.x uses badger@v2, v1 uses badger@v3)?** My idea is to add a new `pkg/migration` with its own go.mod/go.sum in charge of reading all the data from the old version of the datadir - therefore isolating all the old deps in this external package. Practically, this would be a merge of the old `repoManager`, `pubsub` and `macaroon` services (did I forget some?) that write to the datadir. Within `cmd/migration`, instead, we would translate the structs coming from this external pkg to the new format, prepare the new datadirs and flush all the old content (everything's been read, not needed anymore).
- **Why not allowing migrations from `0.x` to `v1`?** Because every bump of version (from `0.x` to `0.y`) coincides with a breaking change, event at the domain level. `v0.9.1` is the last "old" version of the daemon, meaning that versions after that one are v1 or higher, so the operator must make sure to run exactly `v0.9.1` to migrate to `v1`.

sekulicd:
This contains working code for migrating v091 Vault to v1 Ocean's Wallet.
This uses V2 badger/badgerhold to fetch v091 Vault and V3 badger/badgerhold to insert Wallet into Ocean datadir db.
This adds migration for trade domain.
This adds migration for deposits domain.
This adds migration for withdrawals domain.
This adds migration for webhooks.
This adds migration for tls.
This adds migration for macaroons.

Closes #715.

@altafan please review.


cmd/migration/v0.9.1-v1/v091-domain/vault_repository.go
```diff
@@ -0,0 +1,29 @@
...
+func NewVaultRepositoryImpl(store *badgerhold.Store) VaultRepository {
+	return &vaultRepositoryImpl{store}
+}
+
+func (v *vaultRepositoryImpl) GetVault() (*Vault, error) {
+	var vault Vault
+	if err := v.store.Get(vaultKey, &vault); err != nil {
+		if err == badgerhold.ErrNotFound {
+			return nil, nil
+		}
```
altafan: either we remove this or we return a `vault not found` error because we want the vault to be there.. if not let's notify it.

cmd/migration/v0.9.1-v1/mapper/vault_wallet.go
```diff
@@ -0,0 +1,118 @@
...
+			return nil, err
+		}
+
+		accounts[namespace] = &v1domain.Account{
+			AccountInfo: v1domain.AccountInfo{
+				Namespace: namespace,
+				Label:     label,
+				Xpub:      xpub,
+				DerivationPath: fmt.Sprintf(
+					"%s/%d", v091domain.RootPath, v.AccountIndex,
```
altafan: Watch out! Account must be hardened:
```suggestion
					"%s/%d'", v091domain.RootPath, v.AccountIndex,
```

cmd/migration/v0.9.1-v1/mapper/vault_wallet.go
```diff
@@ -0,0 +1,118 @@
...
+		NextAccountIndex:    uint32(highestAccountIndex + 1),
+	}, nil
+}
+
+func (m *mapperService) getLabel(accountIndex int) (string, error) {
+	switch accountIndex {
+	case FeeAccount:
+		return "fee_account", nil
+	case WalletAccount:
+		return "wallet_account", nil
```
altafan: Ah! I didn't think about migrating this account, which is no more used or supported in v1. Anyway, I think it's fine to call it `personal_account`, which clarifies better the purpose of this account, the user will change it later by using the ocean cli eventually:
```suggestion
	case WalletAccount:
		return "personal_account", nil
```

cmd/migration/v0.9.1-v1/mapper/vault_wallet.go
```diff
@@ -0,0 +1,118 @@
...
+	case FeeFragmenterAccount:
+		return "fee_fragmenter_account", nil
+	case MarketFragmenterAccount:
+		return "market_fragmenter_account", nil
+	case UnusedAccount3:
+		return "unused3", nil
```
altafan: This can be removed since account `4` is never used in tdex-daemon v0.x

cmd/migration/v0.9.1-v1/v091-domain/market_domain.go
```diff
@@ -0,0 +1,277 @@
...
+var (
+	ErrInvalidOptsType = errors.New("opts must be of type BalancedReservesOpts")
+	ErrAmountTooLow    = errors.New("provided amount is too low")
+	ErrAmountTooBig    = errors.New("provided amount is too big")
+	ErrBalanceTooLow   = errors.New("reserve balance amount is too low")
+
+	tenThousands = decimal.NewFromInt(10000)
+)
+
+type BalancedReservesOpts struct {
```
altafan: why do we need formulas? isn't it enough to translate the strategy type in accordance with the changes made from v0.x to v1.x without balanced/pluggable implementations?

cmd/migration/v0.9.1-v1/v091-domain/market_domain.go
```diff
@@ -0,0 +1,277 @@
...
+	FixedFee            FixedFee
+	Tradable            bool
+	Strategy            MakingStrategy
+	Price               Prices
+}
+
+func (m Market) AccountName() string {
+	buf, _ := hex.DecodeString(m.BaseAsset + m.QuoteAsset)
+	return hex.EncodeToString(btcutil.Hash160(buf))[:5]
+}
```
altafan: this is a "hack", isn't it? this method does not exist in domainv0, correct? totally fine to be here anyway :)

sekulicd: I understood this is how we populate label in v1?

cmd/migration/v0.9.1-v1/v091-domain/market_domain.go
sekulicd: I dont get how we fetch Market from v091 Db, having Strategy field as interface, unless we add implementation for both Balanced and Pluggable?

altafan: we just want to know what's the strategy type.. these apis should be more than enough for the purpose https://github.com/tdex-network/tdex-daemon/blob/master/internal/core/domain/market_service.go#L46-L53.

Strategy impls would be needed if we have to calculate some preview.. not at all for migration

cmd/migration/v0.9.1-v1/v091-domain/market_domain.go
altafan: yeah it's fine.. i was just saying that this method does not really exist in v0.9 domain... nevermind

altafan (changes requested):
@sekulicd I made a local test and here are some considerations:
- no need for a `--network` flag because we can retrieve this info from the the vault domain (v0).
- we should suppress errors for tls and macaroons in case the folders are not there because the operator might have disabled them.
- webhooks haven't been migrated with error `store is locked`, we should unlock it before performing the migration
- markets haven't been migrated! Not sure if trades have been migrated 'cause there are no markets for which listing trades in my daemon-v1
- there shouldn't be any need to rewrite v1 data structs like trade domain or webhook. We should reuse what we already have within `internal/`
- we should add defaults to the flags so that, without any specific one, the migration runs on the default daemon's datadir. The existing datadir should be renamed to `tdex-daemon-v0` and a new one should take its place along with the `oceand` one. Ideally, after one runs the migration he should be able to run ocean and tdex-daemon without any other intermediate step (if not required because he specifies different folders with flags).

tiero:
> The existing datadir should be renamed to tdex-daemon-v0 and a new one should take its place along with the oceand one. Ideally, after one runs the migration he should be able to run ocean and tdex-daemon without any other intermediate step (if not required because he specifies different folders with flags).

What about migrating to a new folder? `.tdexd` or `.tdex-daemon-v2`

cmd/migration/v0.9.1-v1/main.go
```diff
@@ -0,0 +1,379 @@
...
+	}
+	log.Info("tls migration completed")
+
+	log.Info("macaroons migration started")
+	if err := migrateMacaroons(v091DataDir, v1TdexdDataDir); err != nil {
+		log.Errorf("error while migrating macaroons: %s", err)
+	}
+	log.Info("macaroons migration completed")
+
+	migrateStats()
```
altafan: remove, we don't want to migrate stats

cmd/migration/v0.9.1-v1/mapper/trade.go
```diff
@@ -0,0 +1,84 @@
...
+	market, err := m.v091RepoManager.MarketRepository().GetMarketByAssets(
+		trade.MarketBaseAsset, trade.MarketQuoteAsset,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	return &domain.Trade{
+		Id:               trade.ID.String(),
+		Type:             0, // TODO
```
altafan: this is the trade type, either BUY or SELL

cmd/migration/v0.9.1-v1/mapper/trade.go
```diff
@@ -0,0 +1,84 @@
...
+		MarketName:       market.AccountName(),
+		MarketBaseAsset:  trade.MarketBaseAsset,
+		MarketQuoteAsset: trade.MarketQuoteAsset,
+		MarketPrice: domain.MarketPrice{
+			BasePrice:  trade.MarketPrice.BasePrice.String(),
+			QuotePrice: trade.MarketPrice.QuotePrice.String(),
+		},
+		MarketPercentageFee: domain.MarketFee{
+			BaseAsset:  0, // TODO
+			QuoteAsset: 0, // TODO
```
altafan: in v0 the percentage fee is a single field valid for both base and quote asset, let's use the single v0 value for both fields of v1.

cmd/migration/v0.9.1-v1/mapper/trade.go
```diff
@@ -0,0 +1,84 @@
...
+		},
+		MarketFixedFee: domain.MarketFee{
+			BaseAsset:  uint64(trade.MarketFixedBaseFee),
+			QuoteAsset: uint64(trade.MarketFixedQuoteFee),
+		},
+		FeeAsset:     trade.MarketBaseAsset, // TODO ?
+		FeeAmount:    0,                     // TODO
```
altafan: we have 2 options here: keeping these fields empty or calculating the fee amount always on the base asset and do populate these fields.. I think it's enough for now to leave them empty.

altafan:
> What about migrating to a new folder? .tdexd or .tdex-daemon-v2

I don't really like this solution mainly because the operator would need to customize the datadir of the new version. In addition, in a dockerized environment, one would have to change the volume mounted on the container which doesn't sounds so frictionless to me.

What if we modify the v0.x datadir as the v1 expects, create the ocean folder inside the daemon's datadir and move everything already existing to a dedicated folder like `v0.9.1-datadir`. This way we don't loose anything, and the daemon is still easy to start without modifying anything but the env vars (whether it is served as standalone binary or docker container).

sekulicd:
> > What about migrating to a new folder? .tdexd or .tdex-daemon-v2
> 
> I don't really like this solution mainly because the operator would need to customize the datadir of the new version. In addition, in a dockerized environment, one would have to change the volume mounted on the container which doesn't sounds so frictionless to me.
> 
> What if we modify the v0.x datadir as the v1 expects, create the ocean folder inside the daemon's datadir and move everything already existing to a dedicated folder like `v0.9.1-datadir`. This way we don't loose anything, and the daemon is still easy to start without modifying anything but the env vars (whether it is served as standalone binary or docker container).

This question is related to the deployment strategy and we need to decide exactly how we are going to do this so we know where to put outputs.
From my POV this is one time operational job and i dont think its important that is completely frictionless.
I would maybe force user to provide source datadir, run binary and in same loc two datadirs would be generated(tdexd, ocean), and also maybe we can add docker image that starts all services and binds voluems etc.

cmd/migration/v0.9.1-v1/mapper/trade.go
sekulicd: based on what i can decide if it is trade or sell?

tiero: look from trader standpoint: if a market with ABC as base asset and XYZ as quote asset, when the user is receiveing the ABC (ie. base asset) it's a trade of type BUY, vice versa if its receiving the quote asset is a SELL

tiero:
ok to rename the folder then! I think Docker is where we have more friction anyway, so its important to make a clear migration guide for the raw edges

cmd/migration/v0.9.1-v1/mapper/trade.go
altafan: oh sorry i was strongly convinced the v0 domain stored the trade type but it's really not!
What tiero suggested is ok but I want to add that the v0 domain has a concept of fixed base/quote asset in conifg that means all markets must have the same base/quote asset. Since this info is not stored in the datadir, we may want to add this param as a flag (we should mimic the behavior of the daemon's v0 fixed base asset as default value).

altafan:
My vision is that the dockerized daemon should be migrated by running just one command without any "customization", while for the standalone binary, one would use custom output folders indeed for the new datadirs.

cmd/migration/v0.9.1-v1/mapper/deposit.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	for k, v := range depositsPerTxId {
+		amountPerAsset := make(map[string]uint64)
+		for _, vv := range v {
+			if _, ok := amountPerAsset[vv.Asset]; !ok {
+				amountPerAsset[vv.Asset] = vv.Value
+			} else {
+				amountPerAsset[vv.Asset] += vv.Value
+			}
```
altafan: can be simplified to:
```suggestion
			amountPerAsset[vv.Asset] += vv.Value
```

cmd/migration/v0.9.1-v1/mapper/market.go
```diff
@@ -0,0 +1,47 @@
...
+) (*domain.Market, error) {
+	return &domain.Market{
+		BaseAsset:           market.BaseAsset,
+		QuoteAsset:          market.QuoteAsset,
+		Name:                market.AccountName(),
+		BaseAssetPrecision:  market.BaseAssetPrecision,
+		QuoteAssetPrecision: market.QuoteAssetPrecision,
+		PercentageFee: domain.MarketFee{
+			BaseAsset:  0,
+			QuoteAsset: 0,
```
altafan: these should be set to market v0 percentage fee:
```suggestion
			BaseAsset:  uint64(market.Fee),
			QuoteAsset: uint64(market.Fee),
```

cmd/migration/v0.9.1-v1/mapper/market.go
```diff
@@ -0,0 +1,47 @@
...
+		},
+		FixedFee: domain.MarketFee{
+			BaseAsset:  uint64(market.FixedFee.BaseFee),
+			QuoteAsset: uint64(market.FixedFee.QuoteFee),
+		},
+		Tradable:     market.Tradable,
+		StrategyType: market.Strategy.Type,
+		Price: domain.MarketPrice{
+			BasePrice:  market.Price.BasePrice.String(),
+			QuotePrice: market.Price.QuotePrice.String(),
```
altafan: these fields should be set to empty string `""` in case the v0 version is 0 like:
```
basePrice := ""
if !market.Price.BasePrice.IsZero() {
    basePrice = market.Price.BasePrice.String()
}
```

cmd/migration/v0.9.1-v1/mapper/market.go
```diff
@@ -0,0 +1,47 @@
...
+		PercentageFee: domain.MarketFee{
+			BaseAsset:  0,
+			QuoteAsset: 0,
+		},
+		FixedFee: domain.MarketFee{
+			BaseAsset:  uint64(market.FixedFee.BaseFee),
+			QuoteAsset: uint64(market.FixedFee.QuoteFee),
+		},
+		Tradable:     market.Tradable,
+		StrategyType: market.Strategy.Type,
```
altafan: watch out! strategy type values have changed from [v0](https://github.com/tdex-network/tdex-daemon/blob/master/internal/core/domain/constants.go#L16-L18) to [v1](https://github.com/tdex-network/tdex-daemon/blob/v1/internal/core/domain/constants.go#L4-L7) and must be translated accordingly

cmd/migration/v0.9.1-v1/mapper/trade.go
```diff
@@ -0,0 +1,84 @@
...
+		TraderPubkey: trade.TraderPubkey,
+		Status: domain.TradeStatus{
+			Code:   trade.Status.Code,
```
altafan: watch out! status code values changed from [v0](https://github.com/tdex-network/tdex-daemon/blob/master/internal/core/domain/constants.go#L22-L28) to [v1](https://github.com/tdex-network/tdex-daemon/blob/v1/internal/core/domain/constants.go#L11-L16) and must be translated accordingly

cmd/migration/v0.9.1-v1/mapper/trade.go
```diff
@@ -0,0 +1,84 @@
...
+			Failed: trade.Status.Failed,
+		},
+		PsetBase64:     trade.PsetBase64,
```
altafan: @tiero @sekulicd shall we translate the psetv0 to psetv2? what do you think?

altafan: shall we also translate the swap messages below from proto v1 to proto v2?

cmd/migration/v0.9.1-v1/mapper/unspent_utxo.go
```diff
@@ -0,0 +1,58 @@
...
+		AssetCommitment:     []byte(unspent.AssetCommitment),
+		ValueBlinder:        unspent.ValueBlinder,
+		AssetBlinder:        unspent.AssetBlinder,
+		Script:              unspent.ScriptPubKey,
+		Nonce:               unspent.Nonce,
+		RangeProof:          unspent.RangeProof,
+		SurjectionProof:     unspent.SurjectionProof,
+		AccountName:         market.AccountName(),
+		LockTimestamp:       0,                     //TODO
+		LockExpiryTimestamp: 0,                     //TODO
```
altafan: if the current utxo is locked, let's set lock timestamp as `now` and a lock expiry timestamp of `now + 1 minute`.

BTW, what a crazy operator one would be to do a migration while a trade is ongoing 🤣

cmd/migration/v0.9.1-v1/mapper/unspent_utxo.go
```diff
@@ -0,0 +1,58 @@
...
+		SpentStatus:         v1domain.UtxoStatus{}, //TODO
+		ConfirmedStatus:     v1domain.UtxoStatus{}, //TODO
```
altafan: we can get these info from the explorer.. the v0 domain has just a bool field for spent/confirmed and we have to "enrich" these info manually just in case.

go.mod
```diff
@@ -12,18 +12,22 @@ require (
 	github.com/btcsuite/btcd v0.23.2
 	github.com/btcsuite/btcd/btcec/v2 v2.2.0
 	github.com/btcsuite/btcd/btcutil v1.1.0
+	github.com/btcsuite/btcutil v1.0.3-0.20201208143702-a53e38424cce
```
altafan: we must prevent this, ie. having 2 different versions of btcutil pkg in the main go.mod.

To avoid this, we should isolate the code requiring this old version of btcutil in a dedicated submodule.

cmd/migration/v0.9.1-v1/mapper/trade.go
tiero: what the advantage? we are not going to use those data, as it is there for historical purposes

cmd/migration/v0.9.1-v1/v091-domain/market_domain.go
altafan: I saw that the v0 domain stores the strategy as [`marketmaking.MarketStrategy`](https://github.com/tdex-network/tdex-daemon/blob/master/pkg/marketmaking/market_making.go#L8-L11) so I understood why we need the impls (because of `restoreStrategy`).

since we are not really interested in the formula impl, we could make changes to `NewBalancedStrategy` and `NewPluggableStrategy` factories defined here and just set to formula to `nil` for simplicity.

The best thing though would be isolating this code into a submodule so we can directly reuse pkg/marketmaking from tdex-daemon@v0.9.1. We'll discuss about this offline

go.mod
sekulicd: As discussed to be considered in another PR

cmd/migration/v0.9.1-v1/mapper/unspent_utxo.go
sekulicd: Done

cmd/migration/v0.9.1-v1/mapper/unspent_utxo.go
sekulicd: Done

cmd/migration/v0.9.1-v1/mapper/trade.go
sekulicd: Done

cmd/migration/v0.9.1-v1/mapper/market.go
sekulicd: Done

cmd/migration/v0.9.1-v1/mapper/market.go
sekulicd: Done

cmd/migration/v0.9.1-v1/mapper/market.go
sekulicd: Done

cmd/migration/v0.9.1-v1/mapper/deposit.go
sekulicd: Done

cmd/migration/v0.9.1-v1/mapper/trade.go
sekulicd: Done

cmd/migration/v0.9.1-v1/mapper/trade.go
sekulicd: Done

cmd/migration/v0.9.1-v1/mapper/trade.go
sekulicd: Done

cmd/migration/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+* Standalone binary version
+  ```sh
+  $ migration --password <wallet_password> [--source-version, --v0-datadir, --v1-datadir, --ocean-datadir, --no-backup]
+  ```
+* Docker version
+  ```sh
+  # Pull the latest image
+  $ docker pull ghcr.io/tdex-network/tdexd:latest
+  $ docker run --rm --volume path/to/datadir:/home/tdex/.tdex-daemon --entrypoint 'tdex-migration' ghcr.io/tdex-network/tdexd:latest --password <wallet_password> --/home/tdex/.tdex-daemon/oceand
```
tiero: `--/home/tdex/.tdex-daemon/oceand` are you missing a flag name here?

altafan: yes definitely

altafan (approved):

PR #724, opened 2023-06-21
Hotfix change password

altafan:
Before this there were problems with the macaroon service after changing the main password and unlocking the daemon.

The reason for this bug was due to the fact that the macaroon service was stopped when locking the daemon, making impossible to update its password and therefore unlocking with the new one.

With this, the macaroon service is not stopped anymore when locking the wallet.

Please @sekulicd review this.


sekulicd:
ACK

tiero (approved):

PR #731, opened 2023-08-03
Fix integration test & minor fixes

altafan:
This contains changes for fixing the execution of the e2e on CI.

This also makes changes to the flags defined for `market report` and `market strategy` commands.

Before this, the operator should have specified the time range for `market report` with numeric values, while now more user-friendly flags are specified like `--last-hour`, `--last-day`, etc.
If not specified within the API request, the daemon now makes use of a default time frame that depends on the length of the time range instead of using always `1h` which is ok only for time ranges in the order of few days.

This also adds a new flag `--balanced` to the `market strategy` command so that the operator must specify which strategy he wants for the current market. Before this, instead, only the `--pluggbale` flag was defined.

This also contains fixes to typos and polishes the docker-compose.yml file.

### DIFF
--- a/cmd/tdex/market.go
+++ b/cmd/tdex/market.go
@@ -177,8 +177,13 @@ var (
 		Usage: "updates the current market making strategy, either automated or pluggable market making",
 		Flags: []cli.Flag{
 			&cli.BoolFlag{
+				Name:  "balanced",
+				Usage: "set the strategy to balanced AMM",
+				Value: false,
+			},
+			&cli.BoolFlag{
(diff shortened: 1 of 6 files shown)


tiero:
Have you validated these changes do not breaks the v2 of the dashaboard?

altafan:
These changes do not affect the APIs, rather, they affect only the tdex-cli client, beside adding a default timeframe for market report at app level. The last change might turn out useful for any client in general, but it's still allowed to set a custom frame for the given timerange.

